PR #2, opened 2021-02-03
Event driven compiler

dmdashenkov:
This PR introduces a markup of the event model for the ProtoData compiler.

Events represent the discovery of different elements of a Protobuf abstract syntax tree.

The new `ProtobufSource` projection accumulates data about a Proto file by listening to some of the events.


dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+    SyntaxVersion syntax = 3;
+
+    // File-level options.
+    repeated Option option = 4;
+
+    // The version of the Protobuf syntax, as defined in a Protobuf source file.
+    //
+    enum SyntaxVersion {
+
+        LV_UNDEFINED = 0;
```
armiol, line 68: Not sure why this is `LV_...`. Please elaborate on this.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+        PROTO2 = 1;
+
+        PROTO3 = 2;
+    }
+}
+
+// Name of a Protobuf type.
+message TypeName {
+
+    // The name, including the package.
```
armiol, line 79: I would design this type so that we could introduce more types (or extend this one?) in case we need a "type name that's just a name" and a "FQN type name which has a type URL prefix along with the package and the simple type name". The problem here is to pick proper naming strategy, as we'll not be able to use `TypeName` again :)

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+message MessageType {
+
+    TypeName name = 1;
+
+    // The prefix of the type's URL.
+    //
+    // A type URL has the following structure: "<type URL prefix>/<type FQN>". For more info, see
+    // the definition of `google.protobuf.Any.type_url`.
+    //
+    string type_url_prefix = 2;
```
armiol, line 93: Continuing my previous comment. To me, a type URL seems just as good a part of a name as a type package. But for some reason, the package is "inside" the `TypeName`, and the type URL isn't.
Therefore I'm proposing to think about the `TypeName` API once again.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+    // For example, in a message:
+    // ```
+    // message Foo {
+    //     string bar = 2;
+    //     int32 baz = 1;
+    // }
+    // ```
+    // Field `bar` has the order 0 and the field `baz` has the order 1.
+    //
+    uint32 order_of_declaration =54;
```
armiol, line 187: Please format this one. And, probably, reconsider the `54` (I would understand if it was `42`, but `54` seem misleading).

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+
+    // The flag telling if the field is a list.
+    //
+    // A list is any `repeated` field which is not a `map`.
+    //
+    bool is_list = 6;
+
+    // The flag telling if the field is a map.
+    bool is_map = 7;
```
armiol, line 196: Please consider using a `oneof` for `is_list`, `is_map` and `map_key_type`.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+
+// Name of a group of fields.
+message GroupName {
+
+    // Name of the `oneof` block.
+    string value = 1;
+}
+
+// Group is a set of fields declared within one `oneof` block.
+message Group {
```
armiol, line 222: Please consider renaming this to `OneofGroup` (and `GroupName` to `OneofName`). That would eliminate any potential confusions in the generated code, which would eventually happen with such a widespread word as "group".

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+    GroupName name = 1;
+
+    // Fields declared in this group.
+    repeated Field field = 2;
+}
+
+// A Protobuf option.
+//
+// An option may be declared for a whole file, a type (message, enum, or service), for
+// a message field, a separate rpc, or an enum value.
```
armiol, line 233: did you mean a `standalone RPC service`?

compiler/src/main/proto/spine/protodata/events.proto
```diff
@@ -0,0 +1,116 @@
...
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.protodata";
+option java_outer_classname = "EventsProto";
+option java_multiple_files = true;
+
+import "spine/protodata/ast.proto";
+
+// Emitted when processing of a file begins.
```
armiol, line 40: I would suggest to tell a bit more here (and in the past tense, as well). In particular, there may be many "actors" who process a file. But I guess we mean a particular one, such as `protoc`.

compiler/src/main/proto/spine/protodata/events.proto
```diff
@@ -0,0 +1,116 @@
...
+
+// Emitted when a file-level option is found.
+message FileOptionDiscovered {
+
+    Path file = 1;
+
+    Option option = 2;
+}
+
+// Emitted when processing of a file ends.
```
armiol, line 54: Does it mean that the code generation is done at this point? Maybe, "processing" is just a too generic of a term to be used in similar cases.

You see, with "processing start" the meaning of "processing" does not really matter, as we are clearly still able to supply our 5 cents. But with "processing complete" it seems that there is nothing we can do at this point. While in fact, that's probably just a discovery phase which is complete.

compiler/src/main/proto/spine/protodata/source.proto
```diff
@@ -0,0 +1,52 @@
...
+option java_outer_classname = "ProtobufSourceProto";
+option java_multiple_files = true;
+
+import "spine/protodata/ast.proto";
+
+// A projection of a complete source file.
+//
+// Includes all the message types declared in the file, along with their fields.
+//
+message ProtobufSource {
```
armiol, line 44: As a "Protobuf source" per se may mean a "source of a certain Protobuf message", I would suggest to name this type "ProtoSourceFile".

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,256 @@
...
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.protodata";
+option java_outer_classname = "AstProto";
+option java_multiple_files = true;
+
+import "google/protobuf/any.proto";
+
+// A path in a file system.
+//
+message Path {
```
armiol, line 42: I would explicitly tell that this may refer to any FS, be it a file or a directory.

And one more thing. Seems we are holding to the path separators "as-is"? If so, let's mention that as well.

armiol (changes requested):

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
...
+    `kotlin-dsl`
+}
+
+kotlinDslPluginOptions {
+    experimentalWarning.set(false)
+}
+
+repositories {
+    mavenLocal()
+    jcenter()
```
alexander-yevsyukov, line 37: Please add Maven Central after `jcenter()`.

compiler/src/main/kotlin/io/spine/protodata/AstExtension.kt
```diff
@@ -0,0 +1,46 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata
+
+/**
+ * Obtains the fully qualified name of the type.
+ */
+public fun MessageType.fqn(): String = name.fqn()
```
alexander-yevsyukov, line 32: Let's use `fullName()` instead of `fqn()`.

compiler/src/main/kotlin/io/spine/protodata/AstExtension.kt
```diff
@@ -0,0 +1,46 @@
...
+public fun TypeName.fqn(): String = "${packageName}.${simpleName}"
+
+public fun TypeName.typeUrl(): String = "${typeUrlPrefix}.${fqn()}"
+
+public fun Field.isMap(): Boolean = hasMap()
+
+public fun Field.isList(): Boolean = hasList()
+
+public fun Field.isRepeated(): Boolean = isMap() || isList()
+
```
alexander-yevsyukov, line 45: Please add one-line docs to all of these extensions.

settings.gradle.kts
```diff
@@ -24,6 +24,13 @@
...
 
-
 rootProject.name = "ProtoData"
 
+include("compiler")
+
+dependencyResolutionManagement {
+    repositories {
+        mavenLocal()
+        jcenter()
```
alexander-yevsyukov, line 34: Please add Maven Central too.

alexander-yevsyukov (approved):
LGTM, with a couple of minor requests.

settings.gradle.kts
yuri-sergiichuk: JCenter will be shut down in less than 3 months (see the official announcement [here](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/)). I guess it worth not mentioning it at all.

dmdashenkov:
@armiol, PTAL again.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,287 @@
...
+    string package_name = 2;
+
+    // The version of the Protobuf syntax.
+    SyntaxVersion syntax = 3;
+
+    // File-level options.
+    repeated Option option = 4;
+
+    // The version of the Protobuf syntax, as defined in a Protobuf source file.
+    //
```
alexander-yevsyukov, line 70: Please do not add empty lines after one-line comments.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -0,0 +1,287 @@
...
+        // `repeated` and `map` fields cannot be a part of a `oneof`.
+        //
+        OneofName oneof_name = 9;
+    }
+
+    // Field-level options.
+    repeated Option option = 10;
+
+    // Additional type details for map fields.
+    message MapField {
```
alexander-yevsyukov, line 232: How about naming it `OfMap`, so that the full type name would be `Field.OfMap`?

compiler/src/test/kotlin/ContextTest.kt
```diff
@@ -0,0 +1,39 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import io.spine.protodata.ProtoDataContext
+import io.spine.protodata.ProtoSourceFileProjection
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+
+class `ProtoData context should` {
```
alexander-yevsyukov, line 32: Is it possible to escape a tick symbol in Kotlin strings? E.g. something like <pre>\\`</pre>?

alexander-yevsyukov (changes requested):
Mostly LGTM, please see my comments.

dmdashenkov:

armiol (approved):

alexander-yevsyukov (approved):
LGTM

PR #4, opened 2021-02-15
Compiler events from descriptors

dmdashenkov:
In this PR we introduce the way of producing compiler events from a `FileDescriptorSet`.
It's not meant to be a public API of the tool, but this is how we will publish events.
For emitting events to the `ProtoData` context, we leverage the `ThirdPartyContext` API.

### DIFF
--- a/compiler/src/main/kotlin/io/spine/protodata/CompilerEvents.kt
+++ b/compiler/src/main/kotlin/io/spine/protodata/CompilerEvents.kt
@@ -73,23 +73,34 @@ import io.spine.protodata.PrimitiveType.TYPE_STRING
 import io.spine.protodata.PrimitiveType.TYPE_UINT32
 import io.spine.protodata.PrimitiveType.TYPE_UINT64
 
-internal object Parser {
+/**
+ * A factory for Protobuf compiler events.
+ */
+internal object CompilerEvents {
 
-    fun parse(desc: FileDescriptorSet): Sequence<EventMessage> = produceEvents(desc)
-
-
-    private fun produceEvents(desc: FileDescriptorSet): Sequence<EventMessage> {
+    /**
+     * Produces a sequence of events based on the given descriptor set.
+     *
+     * The sequence is produced lazily. An element is produced only when polled.
+     *
+     * The resulting sequence is always finite, it's limited by the type set.
+     */
+    fun parse(desc: FileDescriptorSet): Sequence<EventMessage> {
         val files = FileSet.of(desc.fileList)
         return sequence {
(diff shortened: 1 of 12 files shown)


dmdashenkov:

compiler/src/main/kotlin/io/spine/protodata/CompilerEvents.kt
```diff
@@ -0,0 +1,397 @@
...
+        val files = FileSet.of(desc.fileList)
+        return sequence {
+            files.files().forEach { produceFileEvents(it) }
+        }
+    }
+
+    /**
+     * Yields compiler events for the given file.
+     *
+     * Opens with an [EnteredFile] event. Then go the events regarding the file metadata. Then go
```
armiol, line 98: I think multiple `Entered...` and `Exited...` events are less convenient in terms of autocompletion and potential keyboard mistype — and less common as to the traditional ES naming — than the approach with the leading subject and trailing participle, such as `FileEntered`.

Please consider switching to a more traditional naming approach.

compiler/src/main/kotlin/io/spine/protodata/Context.kt
```diff
@@ -37,9 +42,35 @@ public object ProtoDataContext {
...
             .singleTenant("ProtoData")
             .add(ProtoSourceFileRepository())
-            .build()
+    }
+}
+
+/**
+ * The `Protobuf Compiler` third-party bounded context.
+ */
+public object CompilerEventsContext {
```
armiol, line 59: Can't we name the object `ProtobufCompilerContext` as well?

compiler/src/main/kotlin/io/spine/protodata/CompilerEvents.kt
```diff
@@ -0,0 +1,397 @@
...
+import io.spine.protodata.PrimitiveType.TYPE_SINT32
+import io.spine.protodata.PrimitiveType.TYPE_SINT64
+import io.spine.protodata.PrimitiveType.TYPE_STRING
+import io.spine.protodata.PrimitiveType.TYPE_UINT32
+import io.spine.protodata.PrimitiveType.TYPE_UINT64
+
+/**
+ * A factory for Protobuf compiler events.
+ */
+internal object CompilerEvents {
```
armiol, line 79: This  yielding approach looks fun to me! Let's try it out.

armiol (commented):
@dmdashenkov LGTM in general. Please see my comments. And let's discuss vocally if needed.

dmdashenkov:
@armiol, PTAL again.

armiol (approved):

PR #5, opened 2021-02-19
Compiler event subscribers

dmdashenkov:
In this PR we introduce the low-level API for `Subscriber`s and `Renderer`s.

A `Subscriber` listens to compiler events and produces `CodeEnhancement`s. A `Renderer` applies `CodeEnhancement`s from many subscribers to source code.


compiler/src/main/kotlin/io/spine/protodata/ReflectiveBuilder.kt
```diff
@@ -0,0 +1,69 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata
+
+import kotlin.reflect.KClass
+import kotlin.reflect.KVisibility
+
+/**
+ * A builder for creating instances of classes defined by a downstream.
```
armiol, line 33: Let's rephrase `by a downstream`. This is way too generic, and therefore, may be confusing.

compiler/src/main/kotlin/io/spine/protodata/ReflectiveBuilder.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Creates an instance of `T`.
+     *
+     * @param className name of the concrete class to instantiate
+     * @param classLoader the [ClassLoader] to load the class by its name
+     */
+    fun createByName(className: String, classLoader: ClassLoader): T {
+        val cls = classLoader.loadClass(className).kotlin
+        @Suppress("UNCHECKED_CAST")
```
armiol, line 45: I would mention in the method-level docs that the caller is responsible for supplying a proper type by its name.

compiler/src/main/kotlin/io/spine/protodata/Main.kt
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Launches the CLI application without finishing the process after execution.
+ */
+@VisibleForTesting
+internal fun launchApp(vararg argv: String) {
+    Run().parse(argv.toList())
+}
+
+/**
+ * The main CLI command which performs the ProtoData code generation tasks.
```
armiol, line 63: Could you please describe the CLI params here?

While I realise those may not be final at this point, it would be nice to have them in order to understand the idea. It's way too long to understand them from the code.

compiler/src/test/kotlin/io/spine/protodata/E2ETest.kt
```diff
@@ -0,0 +1,136 @@
...
+import kotlin.io.path.writeBytes
+import kotlin.io.path.writeText
+import kotlin.reflect.jvm.jvmName
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+
+class `Commad-line app should` {
```
armiol, line 47: `Command`.

compiler/src/main/kotlin/io/spine/protodata/subscriber/Subscriber.kt
```diff
@@ -0,0 +1,98 @@
...
+import io.spine.server.type.EventClass
+import io.spine.server.type.EventEnvelope
+
+/**
+ * A subscriber to a compiler event.
+ *
+ * Instances of `Subscriber`s are created via reflection. It is required that the concrete classes
+ * have a `public` no-argument constructor.
+ */
+public abstract class Subscriber<E : EventMessage>(
```
armiol, line 42: Initially, I thought compiler events were translated into Spine events. And the subscribers were Spine-powered `@Subscriber`s. Now I see you have chosen a different approach. I am not sure why this is so.

Perhaps, we could meet again and draw some things on a board?

compiler/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
```diff
@@ -0,0 +1,74 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata.renderer
+
+import io.spine.protodata.ReflectiveBuilder
+import io.spine.protodata.subscriber.CodeEnhancement
+
+/**
+ * A `Renderer` takes existing source code, modifies it with a number of
```
armiol, line 33: Not sure if you omitted it so far, or missed it. But we are going to need a concept that _creates_ new files, not just modifies the existing source code. It may be this `Renderer` or a whole another type, but still that is going to be the case.

armiol (changes requested):

dmdashenkov:
@armiol, PTAL again.

compiler/src/main/kotlin/io/spine/protodata/subscriber/CodeEnhancement.kt
```diff
@@ -34,5 +34,22 @@ package io.spine.protodata.subscriber
  *
  * This is a marker interface. Concrete implementations must include all the info needed by
  * the renderers to apply the enhancement to source code.
+ *
+ * An enhancement should be self sufficient. This means not relying on other enhancements.
```
armiol, line 38: `self-sufficient`.

armiol (approved):
@dmdashenkov LGTM here and let's discuss the following actions in person. I think there are still several feature-points to be addressed in future PRs.

PR #6, opened 2021-03-12
Improve public API and publish artifacts

dmdashenkov:
In this PR we perform a lot of minor API improvements and integration fixes.

### API simplification

We hereby delete the `Subscriber` and `CodeEnhancement`s and replace them with a simpler flow. Now, we emit the Protobuf compiler events only once. The `Renderer` should query Projections from the `ProtoData` context for all its needs. We also allow, via a new mechanism of `ContextExtension`s to add arbitrary entities to the `ProtoData` context, making room for more complex use cases. For example, users who would like to build their own extendible tool on top of ProtoData will an ability to build public APIs based on event models. One such user is the Spine Validation library.

### Introduction of `OptionsProvider` 

`OptionsProvider` is a new extension point for the tool users. `OptionsProvider`s register Protobuf message extensions, containing custom options, for the tool to pick up.

If a custom option is used in the Protobuf sources but not registered, ProtoData will ignore it, as there is no way to get any meaningful data on that option. The unknown fields mechanism is of little help here, since it provides raw serialized data with no hints regarding the type of the option.

### Publishing

In this PR we also publish the artifacts to the Maven repository.

We choose GitHub Packages for its simplicity on the publishing site and apparent stability.

We publish a separate artifact for each of the subprojects in this repo. The `protoc` artifact is made for passing it to the Protobuf Gradle plugin.

We also publish a separate artifact for the `cli` subproject which contains the whole CLI application. The installing instructions are included in the README.


dmdashenkov:

compiler/src/main/kotlin/io/spine/protodata/ContextExtension.kt
```diff
@@ -24,14 +24,17 @@
...
+ * An extension above the `ProtoData` bounded context.
+ */
+public interface ContextExtension {
 
-    override fun process(event: TypeEntered): Iterable<CodeEnhancement> =
-        listOf(AddUnderscore(event.type.name))
+    /**
+     * Registers the extension entities in the bounded context.
+     */
+    public fun fillIn(context: BoundedContextBuilder)
```
alexander-yevsyukov, line 39: Do we really need the `public` modifier here and above?

compiler/src/main/kotlin/io/spine/protodata/Pipeline.kt
```diff
@@ -31,70 +31,49 @@ import io.spine.base.Production
...
     init {
         val config = ServerEnvironment.`when`(Production::class.java)
-        config.use(PrunableTransport)
         config.use(InMemoryStorageFactory.newInstance())
     }
 
     /**
      * Executes the processing pipeline.
      */
     public operator fun invoke() {
```
alexander-yevsyukov, line 64: Again, I'm not sure we need to add `public` here and above which is default in Kotlin.

alexander-yevsyukov (commented):
Please see my comments. I'm about half way through in the initial scan, and I'm going to have a closer look in IDEA tomorrow.

compiler/src/main/kotlin/io/spine/protodata/ContextExtension.kt
```diff
@@ -24,14 +24,17 @@
...
-import io.spine.protodata.TypeEntered
-import io.spine.protodata.subscriber.CodeEnhancement
-import io.spine.protodata.subscriber.Subscriber
+import io.spine.server.BoundedContextBuilder
 
-class TestSubscriber : Subscriber<TypeEntered>(TypeEntered::class.java) {
+/**
+ * An extension above the `ProtoData` bounded context.
+ */
+public interface ContextExtension {
```
alexander-yevsyukov, line 34: Since this is a part of public API, let's put more attention to the name of this thing. The issue with the current name I see is that it exposes details that we want to hide from a user who's new to the DDD.

Even if this is an extension to the context, it may be exposed under a more common name, like `Plugin` or something.

compiler/src/main/kotlin/io/spine/protodata/ContextExtension.kt
alexander-yevsyukov: > fillIn()

Can we do it another way around, i.e. expose things that need to be registered?

Also, can I remove something from the passed builder?

compiler/src/main/kotlin/io/spine/protodata/ProtoSourceFileProjection.kt
```diff
@@ -35,49 +36,49 @@ import io.spine.server.projection.Projection
 public class ProtoSourceFileProjection
     : Projection<FilePath, ProtobufSourceFile, ProtobufSourceFile.Builder>() {
 
-    @Subscribe(external = true)
-    internal fun on(e: FileEntered) {
+    @Subscribe
+    internal fun on(@External e: FileEntered) {
         builder()
```
alexander-yevsyukov, line 41: Please try the `TransactionalEntity.update()` extension recently introduced in `core-java` (see `TransactionalEntityExtensions.kt`.

compiler/src/test/kotlin/io/spine/protodata/CompilerEventsTest.kt
```diff
@@ -0,0 +1,151 @@
...
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
+import io.spine.base.EventMessage
+import io.spine.option.OptionsProto.REQUIRED_FIELD_NUMBER
+import io.spine.option.OptionsProto.TYPE_URL_PREFIX_FIELD_NUMBER
+import io.spine.protodata.test.DoctorProto
+import io.spine.testing.Correspondences.type
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+class `'CompilerEvents' should` {
```
alexander-yevsyukov, line 42: `internal`?

compiler/src/main/kotlin/io/spine/protodata/Context.kt
```diff
@@ -29,51 +29,27 @@ package io.spine.protodata
...
 import io.spine.base.EventMessage
 import io.spine.core.UserId
-import io.spine.protodata.subscriber.Subscriber
 import io.spine.server.BoundedContext
 import io.spine.server.BoundedContextBuilder
 import io.spine.server.integration.ThirdPartyContext
-import io.spine.server.projection.ProjectionRepository
 
 /**
  * A factory for the `ProtoData` bounded context.
```
alexander-yevsyukov, line 37: I would say that “ProtoData” is a brand. It's a name of a library. We need something closer “to the ground” for a context name. 

Consider an analogy, Microsoft Directory is a tool for managing users, it may have a contexts called Users, Permissions, Organizations.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -67,6 +67,11 @@ private constructor(
         public fun fromCode(path: Path, code: String): SourceFile = SourceFile(code, path)
```
armiol, line 67: This method is now unused.

compiler/src/main/kotlin/io/spine/protodata/AstExtension.kt
```diff
@@ -93,6 +100,79 @@ public fun Field.isRepeated(): Boolean = isMap() || isList()
...
+    } else {
+        declaredIn.javaOuterClassName()
+    }
+    return Paths.get(packageName.replace('.', '/'), "$topLevelClassName.java")
+}
+
+/**
+ * Obtains the full name of the Java class, generated from this message.
+ */
+public fun MessageType.javaClassName(declaredIn: File): String {
```
armiol, line 132: This method isn't used either.

I imagine it may be very well required in future, but let's cover it with some tests in scope of this PR.

compiler/src/main/kotlin/io/spine/protodata/CompilerEvents.kt
```diff
@@ -110,16 +110,20 @@ internal object CompilerEvents {
...
             .build()
-
         yield(
             FileEntered
                 .newBuilder()
                 .setFile(file)
                 .build()
         )
-
+        produceOptionEvents(descriptor.options) {
```
armiol, line 119: In the declaration of the `produceOptionEvents` — as long as we are still at 100 symbol width —we should make
the `fun` declaration fit a single line.

(not able to leave a comment there, so writing it down here)

compiler/src/main/kotlin/io/spine/protodata/ContextExtension.kt
armiol, line 34: I would also document it a bit more. I.e. which purpose it serves etc.

compiler/src/main/kotlin/io/spine/protodata/Pipeline.kt
```diff
@@ -31,70 +31,49 @@ import io.spine.base.Production
...
+ * the context receive them. Then, the [Renderer], which is able to query the states of entities in
+ * the `ProtoData` context, alters the source set. This may include creating new files and/or
+ * modifying existing ones. Lastly, the source set is stored onto the file system.
  */
 public class Pipeline(
-    private val subscribers: List<Subscriber<*>>,
-    private val projections: List<ProjectionRepository<*, *, *>>,
-    private val renderer: (List<CodeEnhancement>) -> Renderer,
+    private val extensions: List<ContextExtension>,
+    private val renderer:  Renderer,
```
armiol, line 51: I am not sure why there is just a single renderer. Imagine, we use ProtoData for Spine's `core-java` where we most likely will generate both Java and Kotlin code on top of the same source set — hence, declare two renderers.

Probably, we'll need to discuss that in person.

compiler/src/main/kotlin/io/spine/protodata/QueryBuilder.kt
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            queries.byIds(type, setOf(id))
+        }
+    }
+}
+
+/**
+ * A query which may yield many entities.
+ */
+public class MulticastQuery<T : EntityState>(
```
armiol, line 86: I think there is a confusion between Multicast-Singlecast queries and multicast-unicast routing we already have in Spine.

To me, multicast is more about transmitting data between equal peers, rather than about the power of the result set  in a method call.

I would have just one Query type, with `execute() : Set<T>` and `findOne(): T` methods. It's either way a developer's decision somewhere — be it on the Query implementation or on the method call. But with "one type-two methods" system (c) we'll have one type less to support.

armiol (changes requested):

compiler/src/main/kotlin/io/spine/protodata/Pipeline.kt
dmdashenkov: As discussed, we use the explicit API mode and always mark `public` stuff `public`.

compiler/src/main/kotlin/io/spine/protodata/ContextExtension.kt
dmdashenkov: A user could technically remove something from the builder. I've added a note to the doc warning users from doing so.

Regarding exposing VS registering. I'd like to keep the options open for the users to add different entities, standalone dispatchers, and Bus filters to the context. I suppose we could provide several methods for exposing each one of those kinds of signal handlers, though I maintain that registering is a bit easier for the user.

I don't suppose that users not familiar with the framework will find this particularly hard, considering that they have to had declared and implemented at least one signal handler at this point.

compiler/src/main/kotlin/io/spine/protodata/ProtoSourceFileProjection.kt
dmdashenkov: Right now, ProtoData uses Spine v1.7.0. We'll migrate to v2.0 as soon as `bootstrap` is ready. And to release `bootstrap`, we need to release a few Spine libraries first (such as `web` and `gcloud`).

compiler/src/test/kotlin/io/spine/protodata/CompilerEventsTest.kt
dmdashenkov: I suggest skipping visibility modifiers for tests, just like we do in Java. I don't suppose anyone would ever try to use a test case class.

README.md
```diff
@@ -2,4 +2,52 @@
...
-The project is under development right now and is __not ready for use__.
+The project is under development right now. Proceed with caution.
+
+## Installation
+### *nix
+
+To install ProtoData on a *nix system:
+
+ 1. [Download](https://github.com/SpineEventEngine/ProtoData/packages/710696) 
+    the `executable-<version>.jar` archive.
```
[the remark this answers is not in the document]
dmdashenkov: As the link points at the right Package, we want to:

 1. Make sure the user sees the correct file.
 2. Be honest about the contents of this archive: it's an executable program.

Full coordinates of the package are `io.spine.protodata:executable`. However, the user would see it in the GitHub UI already, so I skipped this info.

dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL.

armiol (changes requested):
@dmdashenkov @alexander-yevsyukov 

As discussed vocally, we would like to address two problems.

1. `Renderer`s may append their content to the same files in the `SourceSet` thus provoking a complex merge. Instead of allowing renderers to append directly, we want to have an API that would allow to collect the desired "appending"s per file for each `Renderer`.

Renderers may still wish to modify (i.e. edit or delete) some contents of the generated file. In this case, we agreed it would be their own responsibility to parse file and modify accordingly.

Once this is solved, we wish to try having multiple `Renderer`s per `Pipeline`, so that exactly the same projections could be re-used.

This point is to be addressed in scope of this PR.

2. At the moment end-users of the library would need to understand some terms from DDD, slowing their progress towards actually using the library.

We are looking forward to create some intermediate API between ProtoData and Spine, so that the terms such as "Bounded Context" and "Projection" would not appear in the ProtoData's public API.

This point will be addressed in a sub-sequent PR.

compiler/src/test/kotlin/io/spine/protodata/CompilerEventsTest.kt
alexander-yevsyukov: If memory serves, when IDEA creates a test suite in Kotlin it does apply `internal`. In Java, when we skip `public` the class becomes `package-private` and is not visible from outside (e.g. when auto-importing in IDEA).

If we skip this it in Kotlin, and assuming a big number of test suites, it is going to impact performance of the IDE sooner or later.

There is also a discipline aspect of having `internal` applied to tests. The code of tests _is_ equally important as production code. Therefore, it should be treated as such. We should not be sloppy about it.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -75,7 +76,19 @@ private constructor(
     /**
      * Writes the source code into the file on the file system.
      */
-    public fun write(charset: Charset = Charsets.UTF_8) {
-        path.writeText(code, charset, WRITE, TRUNCATE_EXISTING)
+    internal fun write(charset: Charset = Charsets.UTF_8, rootDir: Path? = null) {
+        val targetPath = if (rootDir != null) {
```
alexander-yevsyukov, line 80: How about having an extension function for `Path` which does what this `if` does?

alexander-yevsyukov (commented):
Please see my comments.

alexander-yevsyukov:
> "Bounded Context" and "Projection" would not appear in the ProtoData's public API.

I already commented on this, but want to re-iterate: we need to address the name `ContextExtension` too.

dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL again.

The issue regarding hiding DDD terminology will be addressed later. I suppose, we'll have to discuss it in person once more.

cli/src/main/kotlin/io/spine/protodata/cli/Main.kt
```diff
@@ -33,71 +33,73 @@ import com.github.ajalt.clikt.parameters.options.required
...
- * the `CodeGeneratorRequest.file_to_generate` as loaded from the `--request` parameter. Finally,
- * the code enhancements produced by the subscribers are supplied to the renderer, which applies
- * them to the source set with the root path, supplied in the `--source-root` parameter.
- *
- * Note that the [Renderer] is launched even if no enhancements are produced. In order to skip
- * the rendering step altogether (making the whole CLI invocation no-op), a subscriber should
- * produce the [SkipEverything] enhancement. This is a signal for the app to stop the processing and
- * quit.
+ * The command accepts class names for the service provider interface implementations via the CLI
+ * parameters, such as `--plugin`, `--renderer`, and `--options`. Then, using
```
armiol, line 67: Can there be multiple `--renderer`s?

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
```diff
@@ -0,0 +1,102 @@
...
+
+package io.spine.protodata.language
+
+import io.spine.protodata.file.Glob
+import io.spine.protodata.renderer.SourceSet
+
+/**
+ * A programming language or a specific syntax.
+ *
+ * For example, Java, Python 3.x, etc.
```
armiol, line 35: For the example sake, let's say "Java 11" :) Since the syntax differs.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
```diff
@@ -0,0 +1,102 @@
...
+    filePattern: Glob
+) : Language(name, filePattern) {
+
+    override fun comment(line: String): String = "// $line"
+}
+
+/**
+ * A collection of commonly used [Language]s.
+ */
+public object CommonLanguages {
```
armiol, line 91: Let's mention here somehow, how one should extend the set of supported `Language`s. 

This `object` may make an impression that these are the languages which are supported, and that's it.

Ideally, I would see the language support somewhere else, such as dedicated module per language.

compiler/src/main/kotlin/io/spine/protodata/option/OptionsProvider.kt
```diff
@@ -24,12 +24,17 @@
...
 
 /**
- * A test enhancement which suggests adding the dollar sign to the name of a type.
+ * A service provider interface for custom Protobuf options.
  */
-data class AddDollar(val targetName: TypeName) : CodeEnhancement
+public interface OptionsProvider {
+
+    /**
+     * Registers some the custom options in the given [registry].
```
armiol, line 37: My English fails me. I don't understand this sentence.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -0,0 +1,99 @@
...
+    public val label: String
+
+    /**
+     * Locates the line number where the insertion point should be added.
+     *
+     * An insertion point should only appear once in a file.
+     *
+     * @param
+     *     lines a list of code lines in a source file
+     * @return the line number at which the insertion point should be added
```
armiol, line 46: @alexander-yevsyukov @dmdashenkov we recently had a short discussion on whether we should re-think documenting the `@return`s for Kotlin types. Alexander made a point about following the Kotlin conventions (namely, what they use in Collections) and considering each `@return` value description a sentence and completing it with a period. Please see [this conversation](https://github.com/SpineEventEngine/time/pull/112#discussion_r611079655).

I propose to settle on this approach right away in ProtoData.

@dmdashenkov @alexander-yevsyukov please feel free to comment on this.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    override fun toString(): String = name
+}
+
+/**
+ * A C-like language.
+ *
+ * Supports double-slash comments (`// <comment body>`).
+ */
+public class CLike(
```
armiol, line 80: To me, supporting `//` looks more like a trait than a sub-class (being too generic to my taste). However, I could live with this sub-type for now.

Let's just think how we would support more sophisticated characteristics of language/syntax once we need them (e.g. support of `var`s, which is common for JS, Java + and TS, but not for C).

testutil/src/main/kotlin/io/spine/protodata/test/CatOutOfTheBoxReleaser.kt
```diff
@@ -0,0 +1,47 @@
...
+import io.spine.protodata.renderer.SourceSet
+
+/**
+ * Reads every file in the source set and obtains its code, invoking the insertion point rendering,
+ * which is a lazy process and would not happen unless someone requested the source code.
+ *
+ * Releases the proverbial Schrödinger's cat (insertion points) out of the box by observing
+ * the code.
+ */
+public class CatOutOfTheBoxReleaser : Renderer(supportedLanguages = setOf(CommonLanguages.any)) {
```
armiol, line 40: I would go with `CatOutOfTheBoxEmancipator` :)

compiler/src/main/kotlin/io/spine/protodata/Pipeline.kt
```diff
@@ -31,70 +31,49 @@ import io.spine.base.Production
...
  *
- * The pipeline starts by building a Bounded Context with the supplied subscribers.
- * Then, the Protobuf compiler events are emitted for the subscribers to listen. Subscribers produce
- * [CodeEnhancement]s in response to the events.
- * Then, the [Renderer], based on the generated enhancements, alters the source set. This may
- * include creating new files and/or modifying existing ones.
- * Lastly, the source set is stored onto the file system.
+ * The pipeline starts by building the `Code Generation` bounded context with the supplied
+ * [Plugin]s. Then, the Protobuf compiler events are emitted and the subscribers in
+ * the context receive them. Then, the [Renderer], which is able to query the states of entities in
```
armiol, line 45: We have multiple renderers already :)

armiol (changes requested):
@dmdashenkov pretty much LGTM, good job!

Please see my comments and suggestions.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
alexander-yevsyukov: And have language names in quotes as we pass to the constructor.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
alexander-yevsyukov: Is Kotlin C-like?.. Or is it Pascal-like?..

alexander-yevsyukov: Let's re-visit it in the following PR. Have an issue for this, please.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceSet.kt
```diff
@@ -27,27 +27,155 @@
...
+        this.files.values.forEach { it.attachTo(this) }
+    }
 
     public companion object {
 
         /**
          * Collects a source set from a given root directory.
          */
-        public fun fromContentsOf(directory: Path): SourceSet {
+        @JvmStatic public fun fromContentsOf(directory: Path): SourceSet {
```
alexander-yevsyukov, line 67: Can we have the annotation on the previous line, please?

alexander-yevsyukov (approved):
LGTM, with a few minor requests to address before merging.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
dmdashenkov: @armiol, the point is to show examples of both `A programming language` and `a specific syntax`. In practice, I think, most people will just use `"Java"` etc., but showing an example with a version of a language might help some people navigate corner cases.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
dmdashenkov: Added an issue: #8.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
dmdashenkov: I've added some doc explaining how to create own `Language`s.

Also, this is not really a part of specific language support, rather a shortcut for the users of the languages closest and dearest to us.

dmdashenkov:
@armiol, PTAL.

armiol (approved):
@dmdashenkov LGTM, though I still think we'll need to be more precise with the definition of `Language`. I guess we'll see that right away once start developing the libraries of `Renderer`s for various Java versions.

PR #9, opened 2021-04-16
Nested types and Protobuf doc

dmdashenkov:
This PR extends the feature coverage with:
 1. Nested types.
 2. Enum types.
 3. Protobuf services.
 4. Protobuf documentation.

All of the above are now scanned from the input descriptors, included into the compiler events, and added to the `ProtobufSourceFile` projection.


dmdashenkov:
@armiol, PTAL.

compiler/src/main/proto/spine/protodata/ast.proto
```diff
@@ -292,6 +347,93 @@ message OneofGroup {
...
+    string package_name = 2;
+
+    // The prefix of the type's URL.
+    //
+    // A type URL has the following structure: "<type URL prefix>/<package>.<short name>". For more
+    // info, see the definition of `google.protobuf.Any.type_url`.
+    //
+    // Must not include a training slash ("/").
+    //
+    string type_url_prefix = 4;
```
armiol, line 371: I think it should be `3`.

compiler/src/main/proto/spine/protodata/source.proto
```diff
@@ -49,4 +49,8 @@ message ProtobufSourceFile {
     File file = 2;
 
     map<string, MessageType> type = 3;
```
armiol, line 51: While we are here, please document these `map`s. It may be non-obvious what the keys are.

testutil/src/main/proto/spine/protodata/test/doctor.proto
```diff
@@ -10,9 +10,21 @@ option java_outer_classname = "DoctorProto";
 option java_multiple_files = true;
 
 import "spine/time/time.proto";
+import "spine/people/person_name.proto";
 import "google/protobuf/empty.proto";
+import "google/protobuf/wrappers.proto";
 
+// Detached 1.
+
+// Detached 2.
```
armiol, line 19: I would include multiline comments here and in other cases as well.

testutil/src/main/proto/spine/protodata/test/doctor.proto
```diff
@@ -38,5 +59,28 @@ enum Incarnation {
...
+    };
+
+    rpc rescue_call(stream spine.people.PersonName) returns (Yes);
+
+    rpc where_are_you(google.protobuf.Empty) returns (stream google.protobuf.StringValue);
+
+    rpc which_actor(stream Episode) returns (stream spine.people.PersonName);
+}
+
 message _DeleteMe {
```
armiol, line 85: Please have a comment on this, since someone _is_ going to delete it eventually.

armiol (approved):

At this point we still need to sort two things out:

1. More generic naming for public things currently named after ES/CQRS/DDD terms.
2. An example of usage this library in some well-known case, such as generating the validation code in one of our examples (I believe we discussed `Kanban`?).

Looking forward to these :)

PR #10, opened 2021-04-21
DDD newbie-friendly API

dmdashenkov:
## DDD Aliases

In this PR we introduce concepts of `View` and `Policy`.

A `View` is a decorated Spine `Projection`.
A `Policy` is a decorated Spine `AbstractEventReactor`.

Such aliasing should help developers who are not very familiar with DDD and Spine but want to use ProtoData.
These developers will still have to internalize event processing basics, but, at least, we're not overloading them with complex DDD terms.

Documentation for `View` and `Policy` is written as a guide for how to use these classes in the simplest cases.

### Classpath shenanigans

In this PR we address the issue with loading custom entity and event types. The issue was that those classes would only be accessible via a special class loader. Spine does not support such kind of classpath manipulations, so errors occurred.

Now, the launcher script, which was previously auto-generated by Gradle, is a custom Python script which constructs the classpath for the Java process including the custom user classpath, which is supposed to contain the custom entity and event types.

### `UInt`

We remove the usage of Kotlin's `UInt` for now, as it's not possible to use it in Java yet. This is probably going to change once Kotlin 1.5.0 release (not RC) comes out.


dmdashenkov:

alexander-yevsyukov:
@dmdashenkov, has the `UInt` matter changed in the 1.5.0 release?

compiler/src/main/kotlin/io/spine/protodata/plugin/View.kt
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * A view is constructed from the events obtained from the Protobuf Compiler, or generated by
+ * a [Policy].
+ *
+ * Listen to the events using single parameter methods annotated with
+ * [@Subscribe][io.spine.core.Subscribe]. In these methods, change the state of the view via
+ * the `builder()` method.
+ *
+ * We recommend to use `internal` access for the subscriber methods in Kotlin and package-private —
+ * in Java. The methods are not `private`, as they are invoked by the framework, and no `public`,
```
alexander-yevsyukov, line 51: and _not_ `public`?

compiler/src/main/kotlin/io/spine/protodata/plugin/View.kt
```diff
@@ -0,0 +1,134 @@
...
+ * events from policies should not.  See the whole list of Protobuf compiler events
+ * in `spine/protodata/events.proto`.
+ *
+ * Example:
+ * ```
+ * class MyView : View<FieldName, Field, Field.Builder> {
+ *
+ *     @Subscribe
+ *     internal fun on(event: SomethingHappened) {
+ *         // Change the view via `builder()`.
```
alexander-yevsyukov, line 64: Did you have a chance try the `TransactionalEntity.update()` extension recently introduced in `core-java`? It's declared in the `TransactionalEntityExtensions.kt` under the `server`. A block like this:

```kotlin
@Subscribe
internal fun on(event: SomethingHappened) {
    update {
         // Set props here.
    }
}
```
would look more natural for Kotlin.

compiler/src/test/kotlin/io/spine/protodata/AstTest.kt
```diff
@@ -105,10 +105,29 @@ class `AST extensions should` {
             assertThat(className)
                 .isEqualTo("${TestEnv.packageName}.${TestEnv.outerClassName}$${simpleName}")
         }
+
+        @Test
+        fun enum() {
+            val simpleName = "ExplosiveType"
+            val type = EnumType
+                .newBuilder()
```
alexander-yevsyukov, line 113: We can already have `.newBuilder()` on the same line with type. Latest IDEA versions have a formatting setting that covers that, and this setting is turned on in `config`.

alexander-yevsyukov (changes requested):
Looks really good, but please see my comments and suggestions.

armiol (approved):
@dmdashenkov other than Alexander's comments, this changeset looks good to me. Great work!

compiler/src/test/kotlin/io/spine/protodata/AstTest.kt
dmdashenkov: Chain call alignment does not work for Kotlin yet. We'd have something like
```kotlin
val type = EnumType.newBuilder()
    .setField(..)
    .setOther(..)
```
instead of
```kotlin
val type = EnumType.newBuilder()
                   .setField(..)
                   .setOther(..)
```

I suggest wrapping first method calls until Kotlin code style config catches up with the Java one. If you'd like, let's discuss in person tomorrow.

compiler/src/main/kotlin/io/spine/protodata/plugin/View.kt
dmdashenkov: We're still not at the point when we can release `bootstrap`, as far as I can tell.
Collecting all the right artifacts manually is a non-trivial exercise, which I'd very much like to avoid.
We'll migrate ProtoData to Spine 2.0 and as soon as `bootstrap` is available.

cli/launch.template.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
dmdashenkov: This was initially going to be a pair of shell scripts. However, I found that extracting named arguments without removing all the others (since we want to pass them to Java) is a bit tricky. Furthermore, Bash and Batch scripts have a completely different approaches to argument processing.

As I've tried out the Python version, I found it to be more readable as well. That is, of course, subjective.

As an objective benefit, we have to support only one script file, not two.
Given that most engineers are going to have Python already installed, I think it's OK to run it in Python.

compiler/src/main/kotlin/io/spine/protodata/plugin/Plugin.kt
```diff
@@ -61,6 +69,9 @@ public interface Plugin {
  */
 internal fun BoundedContextBuilder.apply(plugin: Plugin) {
     plugin.views.forEach {
+        add(ViewRepository.default(it))
+    }
+    plugin.viewRepositories.forEach {
```
alexander-yevsyukov, line 74: What happens if we add a `View` and then a custom `Repository` which manages the same kind of view?

compiler/src/main/kotlin/io/spine/protodata/Strings.kt
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Obtains the same string but in camel case instead of snake case.
+ *
+ * Examples:
+ *  - `"aaa"` becomes `"Aaa"`;
+ *  - `"field_name"` becomes `"FieldName"`;
+ *  - `"TypeName"` stays `"TypeName"`;
+ *  - `"___u_ri____"` becomes `"URi"`.
+ */
+@Suppress("FunctionName") // Demonstrates the CamelCase example.
```
alexander-yevsyukov, line 49: Why don't we do it the same way for `TitleCase` then?

alexander-yevsyukov: I think having good KDoc is quite enough. Most people know what a camel case is.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
```diff
@@ -96,13 +96,17 @@ public class CLike(
  */
```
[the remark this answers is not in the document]
alexander-yevsyukov: Maybe, at least, `SlashCommentLanguage` then? After “Fat Bottomed Girls”.

compiler/src/main/kotlin/io/spine/protodata/language/Language.kt
```diff
@@ -90,19 +88,33 @@ public class CLike(
...
+ * [LanguageWithSlashComments].
  */
 public object CommonLanguages {
 
+    /**
+     * Any language will do.
+     *
+     * This instance indicates that any programming language can be accepted.
+     *
+     * Intended to be used for filtering source files by language via file name conventions. If no
```
alexander-yevsyukov, line 102: Please have `If no` starting from the next line. This way it's easier to read.

compiler/src/main/kotlin/io/spine/protodata/plugin/Policy.kt
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * As a rule of thumb, a policy should read:
+ * Whenever <something happens>, then <something else must happen>.
+ *
+ * For example:
+ * Whenever a field option is discovered, a validation rule must be added.
+ *
+ * To implement the policy, declare a method which reacts to an event with an event:
+ * ```kotlin
+ * class MyPolicy {
```
alexander-yevsyukov, line 45: Do we need to document on inheritance here?

compiler/src/main/kotlin/io/spine/protodata/plugin/Policy.kt
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Please note that when reacting on Protobuf Compiler events, one should mark them as
+ * [@External][io.spine.core.External]. See the whole list of Protobuf compiler events
+ * in `spine/protodata/events.proto`.
+ *
+ * *Note.* Often when talking about policies, people imply converting an event into a command, not
+ * an event. This approach seems too complicated to us at this stage, as not many commands will do
+ * anything but produce events with the same information, thus we directly convert between events.
+ */
+public open class Policy : AbstractEventReactor() {
```
alexander-yevsyukov, line 62: What useful would an instance of _this_ class do?

compiler/src/test/kotlin/io/spine/protodata/StringsTest.kt
```diff
@@ -26,22 +26,40 @@
...
 
-/**
- * The repository for the [ProtoSourceFileProjection].
- */
-public class ProtoSourceFileRepository
-    : ProjectionRepository<FilePath, ProtoSourceFileProjection, ProtobufSourceFile>() {
+class `Strings should` {
+
+    @ParameterizedTest
+    @CsvSource("aaa,Aaa", "field_name,Field_name", "TypeName,TypeName", "_uri,_uri")
```
alexander-yevsyukov, line 37: Круто!

alexander-yevsyukov (commented):
Please see my comments and questions.

dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #14, opened 2021-05-26
Java-friendly SPI anatomy

dmdashenkov:
In this PR we remove abstract properties from SPI interfaces and abstract classes and replace them with similarly named methods.

Consider this Java code:
```java
public final class MyPlugin implements Plugin {

    public Set<Class<? extends View>> getViews() { ... }
}
```
VS
```java
public final class MyPlugin implements Plugin {

    public ImmutableSet<Class<? extends View>> views() { ... }
}
```

The former hints at the presence of some kind of a property, which is not an intended effect.
The latter is easier to read.

Additionally, we replace `Set` with `ImmutableSet` so that the user does not think they can modify this set to change the views. In Kotlin this was already expressed by Kotlin `Set` being immutable, but this constraint is lost in translation to Java.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -67,7 +67,7 @@ spinePublishing {
 
 allprojects {
     group = "io.spine.protodata"
-    version = "0.0.9"
+    version = "0.0.10"
 
     repositories.applyStandard()
 }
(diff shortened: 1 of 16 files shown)


dmdashenkov:
@armiol, PTAL.

alexander-yevsyukov:
> In Kotlin this was already expressed by Kotlin Set being immutable, but this constraint is lost in translation to Java.

@dmdashenkov, but when we use `ImmutableSet` from Guava rather than Kotlin's type we reduce the potential portability of the Kotlin code. Can we solve the issue of possible confusion of non-mutability using annotations, or documentation?

PR #15, opened 2021-05-28
Java code generation tools

dmdashenkov:
This PR introduces the `:codegen-java` module, which provides utilities for generating Java code using ProtoData.


codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
```diff
@@ -0,0 +1,386 @@
...
+ * A fully qualified Java class name.
+ *
+ * In Java, a class name is not a valid expression. Use one of the methods of this class to create
+ * an expression from this class name.
+ */
+public data class ClassName
+/**
+ * Creates a class name from the given FQN.
+ */
+constructor(private val className: String) {
```
alexander-yevsyukov, line 121: Is this layout intended?

dmdashenkov: Yes, primary constructors in Kotlin are declared outside the class body. I was not able to find an official recommendation on this one. 

This alignment is what IDEA does when declaring the primary constructor on a new line.

Some people prefer keeping the `constructor` keyword on the same line with the class name like so:
```kotlin
public class Expression constructor(
    private val className: String
) // ...
```
However, given the presence of a doc, which separates the constructor from the class name, we have no choice but to write it on a new line.
Also, I'd argue, this layout more clear when there is a `constructor` keyword, like in this file.

Without the keyword, the list of constructor params should probably just be written in a column, or on the same line as the class name in some simple cases. For example:
```kotlin
public class Expression(
    private val className: String
) {
    // Class body.
}

public data class Email(val value: String) // No class body, no alignment-based highlithing is required.
```

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
```diff
@@ -0,0 +1,386 @@
...
+        if (code != other.code) return false
+        return true
+    }
+
+    override fun hashCode(): Int =
+        code.hashCode()
+}
+
+/**
+ * An literal `null` expression.
```
armiol, line 83: `A literal`.

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
```diff
@@ -0,0 +1,386 @@
...
+/**
+ * An expression of a Java method call.
+ *
+ * Can be a static or an instance method. In the case of the former, the receiver expression
+ * is a class name. In the case of the latter — the receiver object.
+ */
+public class MethodCall
+@JvmOverloads
+constructor(
+    receiver: Expression,
```
armiol, line 295: I still don't get what a `receiver` is. To me, `receiver` is the one who obtains the result of the expression evaluation. 

Isn't it something like `owner`, in whose context the call is performed? 

I mean, this name either requires more explanation (maybe, examples?). Or, this thing should be named differently.

armiol (approved):
@dmdashenkov LGMT with two comments.

PR #16, opened 2021-06-01
Rollback to Kotlin sets

dmdashenkov:
In a recent PR, we've migrated from using Kotlin `Set`s to using Guava `ImmutableSet`s. The motivation for this change was to bring down the chances of an error for users of ProtoData who write in Java.

However, we also want to untie ourselves from Java and, possibly, with time, make ProtoData multiplatform, in the Kotlin-multiplatform way.
This goal is far away and might and might not change in future, but, for now, we're going to make small steps towards it and see what's what.

Due to the stated goal, we remove usages of Guava immutable collections from ProtoData API.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -68,7 +68,7 @@ spinePublishing {
 
 allprojects {
     group = "io.spine.protodata"
-    version = "0.0.11"
+    version = "0.0.12"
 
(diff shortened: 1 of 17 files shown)


dmdashenkov:

armiol (approved):

PR #21, opened 2021-06-08
Plugin publishing

dmdashenkov:
In this PR we add a missing piece in the Gradle plugin publishing configuration.

Previously, the plugin was not published. It turned out that task `publishPlugins` does not automatically depend on `publish`. In this PR we fix that by manually adding the task dependency.


alexander-yevsyukov (approved):
LGTM

PR #23, opened 2021-06-10
Support for multiple Gradle source sets

dmdashenkov:
In this PR we allow the Gradle plugin to run on any number of source sets. Each declared source set has its own ProtoData task, which doesn't interfere with other ProtoData tasks.

We also add integration tests, which check that the Gradle plugin launches ProtoData and code generation executes correctly.

Lastly, we smooth out public API by removing the `do` prefix from the `doRender(..)` method (now just `render(..)`) and by adding `JavaRenderer`, which helps users map Protobuf model to the Java model.


dmdashenkov:
@armiol, PTAL.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/LaunchProtoData.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.gradle.api.file.Directory
+import org.gradle.api.file.RegularFile
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.Exec
+import org.gradle.api.tasks.Internal
+
+/**
+ * A task which executes a single ProtoData command.
+ *
+ * This class is public to allow users configure dependencies, disable, or configure lazy execution
```
armiol, line 38: I assume you mean "through Gradle configuration" here. And not — what one may think of — that the part of the props may be overridden programmatically, and the rest of them may not.

Also, I would move "Users should NOT ..." to a new paragraph. It's much more visible this way.

README.md
```diff
@@ -32,7 +32,6 @@ To install ProtoData on a *nix system:
  ```
  ./target/dir/install.sh $HOME/custom/installation/dir
  ```
- In any case, the directory must exist. Otherwise, `install.sh` will exit with an error.
 
  4. If you use either `bash` or `zsh` shell, you will be asked to run a `source` command to update
```
armiol, line 36: I thought this file would reflect the support of multiple source sets. I would want to tell end-users how to use ProtoData in their tests, for instance.

tests/consumer/src/main/proto/protodata/test/project.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
```
armiol, line 1: I think it would be interesting (and maybe helpful, too) to put this Proto file — or another one, but similar to this one — to `src/test/proto`. And cover the "multiple source sets" feature as well.

tests/protodata-extension/src/main/java/io/spine/protodata/test/UuidJavaRenderer.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.UUID;
+
+import static java.lang.String.format;
+import static java.lang.System.lineSeparator;
+
+/**
+ * A renderer which adds the {@code randomId()} factory methods to the UUID types.
+ *
+ * <p>A UUID type is a message which only has one field — a {@code string} field called
+ * {@code uuid}.
```
armiol, line 48: Let's invite "called" from the previous line to spend some quality time on this line along with `uuid`.

tests/protodata-extension/src/main/java/io/spine/protodata/test/UuidTypeRepository.java
```diff
@@ -0,0 +1,50 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+import static io.spine.server.route.EventRoute.withId;
+
+/**
+ * The repository for the {@link UuidType} views.
+ *
+ * <p>Configures routing for {@code FieldEntered} events.
+ */
+final class UuidTypeRepository extends ViewRepository<TypeName, UuidTypeView, UuidType> {
```
armiol, line 42: Eventually, it would be good to get rid of repositories at all.

tests/protodata-extension/src/main/java/io/spine/protodata/test/ClassScope.java
```diff
@@ -0,0 +1,95 @@
...
+        typeName = checkNotNull(name);
+    }
+
+    @Override
+    public String getLabel() {
+        return format("class_scope:%s", typeUrl(typeName));
+    }
+
+    @Override
+    public LineNumber locate(List<String> lines) {
```
armiol, line 63: As we most likely will be using this whole test extension as a reference, I propose to document methods like this.

tests/build.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+
+import com.google.protobuf.gradle.protobuf
+import com.google.protobuf.gradle.protoc
+import io.spine.internal.dependency.JUnit
+import io.spine.internal.dependency.Protobuf
+import io.spine.internal.dependency.Truth
+import io.spine.internal.gradle.Scripts
+import io.spine.internal.gradle.applyStandard
+
+@Suppress("RemoveRedundantQualifierName")
```
armiol, line 35: Is it possible to tune IDEA to turn off this warning for `plugins`. It's just we do that in each build :)

tests/build.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+
+        JUnit.api.forEach { testImplementation(it) }
+        Truth.libs.forEach { testImplementation(it) }
+        testRuntimeOnly(JUnit.runner)
+    }
+
+    tasks.withType<Test> {
+        useJUnitPlatform()
+    }
+}
```
armiol, line 79: I feel like this `tests` sub-project lacks a `README.md`. The flow is far from trivial, so a description would become handy for anyone testing their own extensions.

armiol (changes requested):

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/Plugin.kt
```diff
@@ -75,160 +72,48 @@ public class Plugin : GradlePlugin<Project> {
         val extension = Extension(target)
         target.extensions.add("protoData", extension)
         val protoDataConfiguration = target.configurations.create("protoData")
-        createLaunchTask(target, extension, protoDataConfiguration)
+        target.sourceSets.forEach { sourceSet ->
+            val launch = createLaunchTask(target, extension, protoDataConfiguration, sourceSet)
+            target.javaCompileFor(sourceSet)?.dependsOn(launch)
+        }
 
         val resource = Plugin::class.java.classLoader.getResource("version.txt")!!
```
alexander-yevsyukov, line 80: Please have `version.txt` defined as a constant. It's used at least twice in the code.

alexander-yevsyukov (changes requested):
Please see my comments (24/135).

README.md
dmdashenkov: Added a description of using ProtoData on main/test scopes.

armiol (approved):
@dmdashenkov LGTM with a comment.

gradle-plugin/src/test/kotlin/io/spine/protodata/gradle/ExtensionTest.kt
```diff
@@ -82,18 +85,23 @@ class `Plugin extension should` {
...
+    fun `produce source directory`() {
+        val basePath = "my/path"
+        val subDir = "foobar"
+
+        extension.srcBaseDir = basePath
+        extension.srcSubDir = subDir
+
+        val sourceDir = extension.sourceDir(project.sourceSets.getByName(MAIN_SOURCE_SET_NAME))
+        assertThat(sourceDir.get().asFile.toPath())
+            .isEqualTo(project.projectDir.toPath() / basePath / MAIN_SOURCE_SET_NAME / subDir)
```
alexander-yevsyukov, line 105: This is fun!

alexander-yevsyukov (approved):
LGTM

gradle-plugin/src/test/kotlin/io/spine/protodata/gradle/ExtensionTest.kt
dmdashenkov: Inspired by Python, as far as I can tell.

PR #25, opened 2021-06-30
Make `CodeGenerationContext` public for testing

dmdashenkov:
In order to allow users who extend ProtoData to test their code via Spine `BlackBox`, we expose `CodeGenerationContext.builder()` into the public API.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #26, opened 2021-07-06
Generate files into a new dir

dmdashenkov:
In this PR we add a capability for ProtoData to read source files from one directory and write altered (and new) files into another directory.

The source directory will be kept untouched while all the files will be transferred into the target directory, including files that ProtoData does not change at all.

If source and target directories coincide, the changes are applied in-place.

## CLI

In CLI, the `--target-root` option controls the target directory.
If no option is specified, the changes are applied in-place.
If `--source-root` and `--target-root` point at the same place in the file system, changes are applied in-place as well.

Note that the target directory must be an empty directory or not to exist at all. If the target directory path leads to a file or a non-empty directory, an error occurs. The exception is when the target directory is the same as the source directory.

## Gradle

In Gradle, writing ProtoData sources into a separate directory is now the default behaviour.
This enables us to skip ProtoData invocations when no code generation is needed (i.e. perform up-to-date checks).

Also, the ProtoData Gradle plugin now overrides the location where Protoc places files after generation. By default, they now go in a directory under the build dir. And the files generated by ProtoData go under `$projectDir/generated/`.


dmdashenkov:
@armiol, PTAL.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -145,6 +157,14 @@ private constructor(
         }
     }
 
+    /**
+     * Deletes this source file from the file system.
+     */
+    internal fun rm(rootDir: Path) {
```
armiol, line 163: Please document the parameter. It looks confusing, given that _this_ file only resides in a single place on the current-generation FS`es.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceSet.kt
```diff
@@ -46,11 +50,20 @@ internal constructor(
      *
      * Paths of the files must be either absolute or relative to this directory.
      */
-    internal val rootDir: Path
+    private val sourceRoot: Path,
+
+    /**
+     * A directory where the source set should be placed after code generation.
+     *
+     * If same as the `sourceRoot`, all files will be overridden.
```
armiol, line 58: "overwritten"?

compiler/src/test/kotlin/io/spine/protodata/PipelineTest.kt
```diff
@@ -203,6 +205,47 @@ class `'Pipeline' should` {
             .doesNotContain(GenericInsertionPoint.OUTSIDE_FILE.codeLine)
     }
 
+    @Test
+    fun `write code into different destination`() {
+        val destination = Files.createTempDirectory("destination")
```
armiol, line 210: Let's extract this directory creation into a method and move the comment there — instead of having it twice (and in future, maybe even more times) in the codebase.

gradle-plugin/README.md
```diff
@@ -116,6 +116,13 @@ tasks.withType<LaunchProtoData> {
 }
 ```
 
+## Generated files
+
+The plugin changes the configuration of the Protobuf Gradle plugin in such a way that the files
+generated from Protobuf are placed under the `build` directory. After the files are than processed
+by ProtoData, they will turn up in ProtoData target directory, which is, by default,
```
armiol, line 123: Did you mean "end up" or "turn out"?

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/Plugin.kt
```diff
@@ -165,7 +181,24 @@ private fun Project.configureProtobufPlugin(extension: Extension, version: Strin
...
+            sourceSet.java.srcDir(extension.targetDir(sourceSet))
+
+            val sourceDir = file(extension.sourceDir(sourceSet))
+            val task = javaCompileFor(sourceSet)!!
+            task.source = task.source.filter { file -> !file.liesInside(sourceDir) }.asFileTree
+        }
+    }
+}
+
+private fun File.liesInside(directory: File): Boolean =
```
armiol, line 203: I would use "laysInside" to avoid confusion. It's technically the same, but at the same time, it looks truthfully.

Or we could go for "residesIn"?

armiol (approved):
@dmdashenkov LGTM in general. Please see my comments though.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/Plugin.kt
armiol: Reconsidering the "laysInside" option — it's not quite good for this case.

Let's go with something like `residesIn`.

PR #27, opened 2021-07-08
Supply options by file name

dmdashenkov:
In this PR we provide the user with the ability to specify options via just the Protobuf file name.

Previously, we required an `OptionProvider` for all custom options.
Now, in most cases, the user can simply provide the name of the Protobuf file where the options are declared.

CLI example:
```
protodata \
    --options acme/foo/my_options.proto \
    --renderer com.acme.FooRenderer \
    --plugin com.acme.FooPlugin \
    --request ./request_file.bin \
    --target-root ./generated/
```

Gradle example:
```kotlin
protoData {
    options("acme/foo/my_options.proto")

    plugins("com.acme.FooPlugin")
    renderers("com.acme.FooRenderer")
}
```

## Also in this PR

Also, in this PR, we configure the Gradle IDEA plugin so that it does not report "duplicate classes".


dmdashenkov:
@armiol, PTAL.

cli/src/main/kotlin/io/spine/protodata/cli/FileOptionsProvider.kt
```diff
@@ -0,0 +1,72 @@
...
+        val outerClassName = outerClassName()
+        val classLoader = this.javaClass.classLoader
+        val optionsClass = classLoader.loadClass(outerClassName)
+        val method = optionsClass.getDeclaredMethod(
+            "registerAllExtensions", ExtensionRegistry::class.java
+        )
+        method.invoke(null, registry)
+    }
+
+    private fun outerClassName(): String {
```
armiol, line 58: This method reminds me of `io.spine.code.java.ClassName.outerClass(FileDescriptor) : String`.

Any chance we could use that or some other part of what we already have in `base`?

cli/src/test/kotlin/io/spine/protodata/cli/given/CustomOptionRenderer.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.spine.protodata.language.CommonLanguages
+import io.spine.protodata.renderer.Renderer
+import io.spine.protodata.renderer.SourceSet
+import io.spine.protodata.select
+import kotlin.io.path.Path
+
+class CustomOptionRenderer : Renderer(CommonLanguages.any) {
+
+    companion object {
+
```
armiol, line 39: Please kill this line.

armiol (approved):
@dmdashenkov LGTM with some comments to consider.

PR #28, opened 2021-07-09
Skip installation when using Gradle plugin

dmdashenkov:
In this PR we improve the implementation of the Gradle plugin so that the ProtoData CLI tool does not have to be installed on the user's machine.

There are no changes in the pubic API of the plugin except that task `installProtoData` does not exist anymore.

Instead, we fetch the executable JAR from the Maven repository and launch it via a `JavaExec` task.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #30, opened 2021-07-13
Configuration API

dmdashenkov:
## Configuration API

In this PR we introduce configuration API for users to pass custom configuration to their ProtoData extensions.

As of now, users may choose between 5 formats to provide configuration:
 - a Protobuf message encoded in binary;
 - a Protobuf message encoded as a Protobuf JSON string;
 - a plain JSON;
 - a YAML;
 - a plain string value.

JSON and YAML values are deserialized into Java objects using Jackson. Users may choose to use Jackson annotations, modules, etc. to improve the configuration model.

To specify the configuration, users may supply a file with the serialized configuration in one of the supported formats. ProtoData infers the format from the file extension: `.bin` and `.pb` are for Protobuf binary, `.pb.json` is for Protobuf JSON, `.json` is for plain JSON, `.yaml` and `.yml` are for YAML. Users cannot provide a plain string configuration via a file.

Another way to specify the configuration is by providing its raw value via a CLI argument. This option is not available for the Protobuf binary format. A second CLI argument is required to specify the format of the configuration.
For example:
```
protodata <...> --configuration-value "foo:bar" ----configuration-format plain
```

When using the Gradle plugin, the only option for specifying the configuration is by a configuration file.

## Code coverage

In this PR we enable code coverage reports for the project.
We collect coverage info from a selected subset of subprojects.
Some low-hanging-fruit coverage wholes were patched right away. More may come in the nearest future.

### DIFF
--- a/compiler/src/main/kotlin/io/spine/protodata/config/ConfigurationParser.kt
+++ b/compiler/src/main/kotlin/io/spine/protodata/config/ConfigurationParser.kt
@@ -64,16 +64,14 @@ private sealed class ProtobufParser : ConfigurationParser {
         if (!Message::class.java.isAssignableFrom(cls)) {
             throw IllegalStateException("Expected a message class but got `${cls.canonicalName}`.")
         }
-        return doParse(source, cls)
+        @Suppress("UNCHECKED_CAST")
+        return doParse(source, cls as Class<out Message>) as T
     }
 
-    abstract fun <T> doParse(source: ByteSource, cls: Class<T>): T
-
-    @Suppress("UNCHECKED_CAST")
-    protected fun <T> Message.asT() = this as T
-
-    @Suppress("UNCHECKED_CAST")
-    protected fun Class<*>.asMessageClass() = this as Class<out Message>
+    /**
+     * Deserializes the given bytes into a message with the given class.
+     */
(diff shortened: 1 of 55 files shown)


dmdashenkov:
@armiol, PTAL.

compiler/src/main/kotlin/io/spine/protodata/config/ConfigurationParser.kt
```diff
@@ -0,0 +1,161 @@
...
+private sealed class ProtobufParser : ConfigurationParser {
+
+    final override fun <T> parse(source: ByteSource, cls: Class<T>): T {
+        if (!Message::class.java.isAssignableFrom(cls)) {
+            throw IllegalStateException("Expected a message class but got `${cls.canonicalName}`.")
+        }
+        return doParse(source, cls)
+    }
+
+    abstract fun <T> doParse(source: ByteSource, cls: Class<T>): T
```
armiol, line 70: Please document this one.

armiol (approved):
@armiol LGTM with a minor comment.

Also, it would be nice to have a way to pass a plain string as named parameter. We could call it a "command-line"
```
... --param-command-line=vasya:123111
```

(probably, meaning that "vasya" has this password).

PR #31, opened 2021-07-16
Remove Kotlin plugin from a non-Kotlin project

dmdashenkov:
We used to have to apply the Kotlin Gradle plugin to Java-only projects so that IntelliJ does not report false compiler errors.
Now, this bug is fixed in the IDE and we don't have to have the unnecessary plugin there.


alexander-yevsyukov (approved):
LGTM

PR #32, opened 2021-07-29
Use Protobuf compiler insertion points

dmdashenkov:
In this PR we allow the users to base their code generation on the insertion points created by the Protobuf compiler.

Previously, to use those points, users had to define a ProtoData insertion point to mirror the native insertion point.
Now we can just find the native insertion point in a code file.


dmdashenkov:
@armiol, PTAL.

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/ProtocInsertionPoints.kt
```diff
@@ -0,0 +1,94 @@
...
+     * Remember to add a comma after the interface name when using this insertion point.
+     */
+    MESSAGE_IMPLEMENTS,
+
+    /**
+     * The place where implemented interfaces go in a builder class.
+     *
+     * Only available for message types.
+     *
+     * Remember to add a comma after the interface name when using this insertion point.
```
armiol, line 75: An example for these repetitive instructions would be nice.

armiol (approved):
@dmdashenkov LGTM with a comment.

PR #33, opened 2021-08-02
Simplify field getters

dmdashenkov:
Previously, a field getter for a list or a map field used to copy the field contents into a Guava immutable collection.

Now, we remove copying. API users should do it manually if then need.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #35, opened 2021-08-09
Fix qualified type name construction

dmdashenkov:
In this PR we add the missing support for nested type names.

Previously, when printing a name of a nested type, we build an invalid name—without the nesting types.

Also, in this PR we migrate ProtoData to the new Spine and the new Gradle.
Troubles related to GitHub Packages make us disable integration tests for the Gradle plugin for the time being.


dmdashenkov:
@armiol, PTAL.
Most of the changes are copies of the updated `config`.

armiol (approved):

PR #36, opened 2021-08-10
Policy Contract

dmdashenkov:
In this PR we update the declaration of `Policy` to adhere to the newly added to Spine contract method definition.

We also fix an issue with constructing a qualified name of a Protobuf type and address the new deprecation warnings that appeared after migrating to Gradle v7.


dmdashenkov:
@armiol, PTAL.

PR #37, opened 2021-08-11
Fix `@ContractFor` in `Policy`

dmdashenkov:
The annotation on `Policy.whenever` had a wrong type of handler method annotation and thus did not serve its purpose to validate the `protected` method access.


dmdashenkov:
@armiol, PTAL.

PR #40, opened 2021-09-07
Update Spine

dmdashenkov:
In this PR we:
 1. Update Spine versions. Now, the `spine/options.proto` contains the `compare_by` option. After ProtoData includes this option, we can start developing code generation based on `compare_by` in `base`.
 2. Add a convenience method for creating instances of `Just`.
 3. Use `protected` modifier with `Policy.whenever(..)` to comply with the contract method spec.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
@@ -26,8 +26,11 @@
 
 package io.spine.internal.gradle
 
+import com.google.auth.oauth2.GoogleCredentials
+import com.google.cloud.artifactregistry.auth.DefaultCredentialProvider
 import io.spine.internal.gradle.PublishingRepos.gitHub
 import java.io.File
+import java.io.IOException
 import java.net.URI
 import java.util.*
 import org.gradle.api.Project
@@ -96,6 +99,8 @@ data class Credentials(
  */
 object PublishingRepos {
 
+    private const val CLOUD_ARTIFACT_REGISTRY = "https://europe-maven.pkg.dev/spine-event-engine"
+
     @Suppress("HttpUrlsUsage") // HTTPS is not supported by this repository.
     val mavenTeamDev = Repository(
         name = "maven.teamdev.com",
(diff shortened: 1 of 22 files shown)


compiler/src/test/kotlin/io/spine/protodata/plugin/JustTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        val just = Just.just(validEvent)
+        assertThat(just)
+            .containsExactly(validEvent)
+    }
+}
+
+private val validEvent = ProjectCreated
+    .newBuilder()
+    .setId(newUuid())
+    .build()
```
yuri-sergiichuk, line 54: Why not `vBuild` or `buildPartial`?
```suggestion
private val validEvent = ProjectCreated.newBuilder()
    .setId(newUuid())
    .build()
```

yuri-sergiichuk (approved):

buildSrc/src/main/kotlin/io/spine/internal/gradle/Repositories.kt
```diff
@@ -110,6 +115,44 @@ object PublishingRepos {
...
+     * Then, the key must be placed somewhere on the file system and the environment variable
+     * `GOOGLE_APPLICATION_CREDENTIALS` must be set to point at the key file.
+     * Once these preconditions are met, publishing becomes possible.
+     *
+     * ## Implementation note
+     * Google provides [tools](https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools)
+     * for configuring authentication for the Maven repositories, including a Gradle plugin.
+     * However, the plugin is incompatible with Gradle 7.x at the moment.
+     * For now, we reproduce what the plugin does manually. This makes the whole `buildSrc`
+     * depend on the `artifactregistry-auth-common` artifact. Track [this issue](https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools/issues/52)
```
alexander-yevsyukov, line 133: Please have `Track this issue...`  coming as a next paragraph. And “‘please‘ would be nice”.

alexander-yevsyukov (approved):
LGTM, with a couple of minor doc-related suggestions.

compiler/src/test/kotlin/io/spine/protodata/plugin/JustTest.kt
dmdashenkov: I've changed the method to `vBuild()` for now. However, please keep in mind that in future, `vBuild()`-s will die off and `build()` will validate messages.

PR #46, opened 2021-11-10
Update option definitions

dmdashenkov:
We update the Spine dependencies to obtain the new definitions of validation options.


dmdashenkov:
@armiol, PTAL.
Note that most of the changes are just the updates to `buildSrc` (copied twice, in the root project and in the `tests`).

armiol (approved):

PR #47, opened 2021-12-16
Avoid circular `compiler` dependency. Step 2.

alexander-yevsyukov:
This PR continues addressing the circular `:compiler` dependency issue, resolution of which was started in #48: 
 * Forcing versions was temporarily switched off to produce ProtoData artifacts that do not require `:compiler` artifact in `protoData` configuration.
 * Dependency on `mc-java` was temporarily rolled back to pre-ProtoData version.
 * Tests of ProtoData Gradle Plugin were temporarily switched off.


alexander-yevsyukov:
Closing in favor of #49.

PR #48, opened 2021-12-20
Resolve `:compiler` dependency conflict

dmdashenkov:
In this PR we exclude the `:compiler` module dependency from the `protoData` configuration.

This helps us to remove a duplicate `:compiler` artifact (which is not supposed to be there at all).


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #49, opened 2021-12-27
Introduce `fat-cli`

alexander-yevsyukov:
This PR continues addressing the circular dependencies occurring when using the version of `mc-java` (which already uses ProtoData) as the ProtoData dependency. This PR is continuation of works started in #48 and #49.

### `fat-cli`
The main change introduced by this PR is `fat-cli` artifact (which is a fat JAR produced by the Gradle Shadow plugin), dependency on which is used in ProtoData modules _instead_ of just `cli` (which is a normal JAR with dependencies on other ProtoData artifacts).

Using `fat-cli` is switched on when the `System` property `spine.internal.protodata.devmode.enabled` is defined. ProtoData build process which uses earlier version of ProtoData (directly or indirectly via `mc-java` should set this property to any value).

### Turning off Validation
Another “hack” which involves `System` properties is disabling Validation library dependencies in `mc-java` which also leads to circular dependency (because Validation uses ProtoData). The flag which turns off the Validation dependency in `mc-java` is called `spine.internal.validation.disabled`, and is turned on in ProtoData modules using `mc-java`. Please see `build-proto-model.kts` script plugin for details.

### Other notable changes
  * The `--options` CLI parameter was removed because it is possible to detect `.proto` files containing Protobuf extension options by analyzing `FileDescriptor`s.
  * `SourceSet` class was renamed to `SourceFileSet` to avoid the confusion with the class from the Gradle API.
  * `--rendered` CLI parameter was made optional because it is the mode in which ProtoData is used when Validation is turned off in `mc-java`.
  * `LaunchProtoData` task was modified to support optional `source` directory input. Such a case is possible when all `protoc` built-ins are turned off, resulting in no source code being generated. In such a mode, `protoc` works only generating descriptor set files. This mode is used in the `bootstrap` Gradle plugin.
  * `OptionsProvider.dumpTo(ExtensionRegistry)` method was renamed to `registerIn()`.

### Further works
The following works are planned in relation to the circular dependency issue and making usage of ProtoData simpler:
  * The `--request` CLI parameter will be replaced with the parameter holding a reference to a directory with descriptor set files. It turned out that composing `CodeGenerationRequest` is not a trivial task, and is hard to master for ProtoData users.
  * Switching to newer version of `mc-java` which is based on the version ProtoData from this or one of the following PRs.
  * Moving `io.spine.protodata.language.Language` class to `tool-base` so that we can use it universally in other tools (e.g. when defining code-generation in `mc-java` or Model Compiler versions for languages other than Java).
  * Renaming the `testutil` module into `test-env` (and presumably renaming the package `io.spine.protodata.test` to `given`) to reduce the confusion with the publishing this module to `mavenLocal()` and not publishing to public repositories.
  * Introducing mechanism for specifying version of an applied plugin for `build.gradle.kts` files stored in resources. And using it for replacing `"0.1.4"` with currently developed version of ProtoData for `gradle-plugin/src/test/resources/`.
  * Supporting the `fat` mode for Validation so that we can use when developing ProtoData and other tools.


cli/src/main/kotlin/io/spine/protodata/cli/Main.kt
```diff
@@ -84,122 +82,155 @@ private fun readVersion(): String {
...
+
+    /** Configuration option names that are used in the help texts of other options. */
+    private object ConfigOpt {
+        const val FILE = "--configuration-file"
+        const val VALUE = "--configuration-value"
+        const val FORMAT = "--configuration-format"
+    }
+
+    /** Abbreviation extension. */
+    private fun String.ti() = trimIndent()
```
dmdashenkov, line 119: It'd be great to check if the compiler treats this thing the same.

I suspect that the Kotlin compiler applies string manipulations such as `trimIndent()` and `trimMargin()` on literals at compile time. I wonder if it's smart enough to do that with `ti()`.

This is not a call for action, just a point of interest.

dmdashenkov, line 119: Please move this:
  a) to the end of the file outside the class scope (as this is clearly not a true class method);
 OR
  b) to another file.

version.gradle.kts
```diff
@@ -24,7 +24,21 @@
...
+val mcJavaVersion: String by extra("2.0.0-SNAPSHOT.83")
+
+val toolBaseVersion: String by extra("2.0.0-SNAPSHOT.85")
+
+// The version of ProtoData being developed.
+val protoDataVersion: String by extra("0.1.4")
+
+/** The version of ProtoData used for developing [protoDataVersion]. */
+val devProtoDataVersion: String by extra("0.1.2")
+
```
dmdashenkov, line 44: Please kill this line.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
```diff
@@ -41,7 +41,7 @@ import kotlin.text.Charsets.UTF_8
 /**
  * A set of source files.
  */
-public class SourceSet
+public class SourceFileSet
```
dmdashenkov, line 44: If this rename is due to a clash with Gradle, I don't think we should sacrifice brevity. Gradle `SourceSet` and ProtoData `SourceSet` live in different contexts and barely (if ever) meet.

If there is another reason, please, tell me.

dmdashenkov (changes requested):

cli/src/main/kotlin/io/spine/protodata/cli/Main.kt
```diff
@@ -234,18 +265,18 @@ internal class Run(version: String) : CliktCommand(
     private fun createRegistry(): ExtensionRegistry {
         val optionsProviders = loadOptions()
         val registry = ExtensionRegistry.newInstance()
-        optionsProviders.forEach { it.dumpTo(registry) }
+        optionsProviders.forEach { it.registerAt(registry) }
```
armiol, line 268: I would say `registerIn`. 

"Register at" would rather point to the place where the action happens, not to the thing in which we register.  See [this conversation](https://ell.stackexchange.com/a/171467) on the subject.

cli/src/test/kotlin/io/spine/protodata/cli/given/TestOptionProvider.kt
```diff
@@ -33,7 +33,7 @@ import io.spine.protodata.option.OptionsProvider
 
 class TestOptionProvider : OptionsProvider {
 
-    override fun dumpTo(registry: ExtensionRegistry) {
+    override fun registerAt(registry: ExtensionRegistry) {
```
armiol, line 36: See my comment above on this matter.

armiol (commented):
@alexander-yevsyukov LGTM in general. Please see my comments.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
alexander-yevsyukov: `SourceSet` from Gradle is a thing bigger than what we have. It's more than just `Set<SourceFile>`. If we “reuse” the name, the user of ProtoData and `tool-base` APIs should be surprised less. Usages of the class are way wider than its package, and chances that would be usage scenarios where `SourceSet` and `SourceFileSet` would be used in a context of a single user class.

cli/src/main/kotlin/io/spine/protodata/cli/Main.kt
alexander-yevsyukov: It's deliberately put in the “context” of the options block and _before_ the usage so that it is clear why we need the method.

alexander-yevsyukov: Assuming that `trimIndent()` itself is done this way...
```kotlin
public fun String.trimIndent(): String = replaceIndent("")
```
I don't think we're at risk.

alexander-yevsyukov: There's little use of this method outside the class. It's just a way to make the options help look a bit more compact.

armiol (approved):

dmdashenkov (approved):

PR #51, opened 2022-01-11
Update copyright notice

dmdashenkov:
In this PR we update the copyright notice to the year 2022. We also patch the Protobuf version to the newest version to resolve a vulnerability.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #58, opened 2022-01-25
Rename setup archive

alexander-yevsyukov:
This PR continues improvements on the naming of things. In particular:

  * The `testuntil` module was renamed to `test-env`. 
  
  * The `publish` task for the `test-env` artifact was disabled on `Project.afterEvaluate` to make sure it's not published to public Maven repositories. As of previous version (`0.1.6`) it was published despite the `enabled` flat set to `false` in `build.gradle.kts` of the module.
  
  *  The distribution archive for ProtoData CLI was renamed to `protodata-setup`. The related task and configuration were renamed accordingly.
  
  * The name of the installation workflow was improved to highlight the name of the OS under which we run it (Ubuntu).

Other changes:
  * Groovy script plugins were remove since they are no longer used.
  * The `LiceseReport` dependency object was removed too (since it cannot be used from `buildSrc/build.gradle.kts` at all), and the only usage was in the removed Groovy scripts.
  * The License Report was updated to version 2.1. The version 2.0 is no longer available from public repos.
  * The `shadow` plugin was updated to `7.1.2`.
  * Dependencies on `base` and `core-java` were also updated to the latest versions.

Further planned actions:
 * Do not publish `test` artifacts in `cli` module.
 * Have `protodata-` prefix for archives under the `protodata/lib/` in the distribution archive (named `protodata-setup.jar`).
 * Use tag replacement in test build scripts (to save on manual replacement when a version advances).


armiol (approved):
LGTM.

PR #61, opened 2022-01-28
Extract API module

alexander-yevsyukov:
This PR extracts the `gradle-api` module to avoid the circular dependency at the usage sites that need to configure ProtoData code generation via the Gradle plugin.

#### Other notable changes
  * Tests based on build scripts stored in resources now employ the tag replacement for defining applying the currently developed version of the Gradle plugin. 
  * The version of the plugin defined in `version.gradle.kts` is now explicitly used in the integration tests. Please see the build scripts under the `tests` directory for details.
  * Objects `CleanTask`, `LaunchTask`, `CodeGeneratorRequestFile`, and `DevMode` were introduced to make these concepts more pronounced in the code.
  * The Gradle plugin ID was updated to avoid the dash in the name: `io.spine.protodata`.
  * The code was updated to use language-related types from `tool-base`.
  * The plugin version is set to `0.2.0` (to celebrate the glorious February of this year).

#### Things left for the following PRs
- [ ] Do not publish `test` artifacts in `cli` module to public repositories.
- [ ] Have `protodata-` prefix for archives under the `protodata/lib/` in the distribution archive (named `protodata-setup.jar`).


alexander-yevsyukov:
@armiol @dmdashenkov please have a look.

gradle-api/src/main/kotlin/io/spine/protodata/gradle/DevMode.kt
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Serves the detection of the special mode when ProtoData is being used for developing itself.
+ */
+public object DevMode {
+
+    /**
+     * If set to any value, enables the development mode for ProtoData.
+     *
+     * Should only be used when developing ProtoData itself.
```
armiol, line 37: I would think that we need it for `validation` as well. So maybe we could say 

"when developing ProtoData itself, or the ProtoData extensions."

@dmdashenkov FYI.

gradle-api/src/main/kotlin/io/spine/protodata/gradle/Names.kt
```diff
@@ -0,0 +1,63 @@
...
+public object Names {
+
+    /**
+     * The resource file containing the version of ProtoData.
+     *
+     * Such a resource name might be duplicated in other places in ProtoData code base.
+     * The reason for this is to avoid creating an extra dependency for the Gradle plugin,
+     * so that the users wouldn't have to declare a custom Maven repository to use the plugin.
+     */
+    public const val VERSION_RESOURCE: String = "version.txt"
```
armiol, line 41: I would suggest to rename it into "protodata.version" in future. 

The reason is that we may be not the only team using English to name their "utility" files with versions.

gradle-api/src/main/kotlin/io/spine/protodata/gradle/ProtoDataApi.kt
```diff
@@ -0,0 +1,101 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata.gradle
+
+/**
+ * Configures code generation process performed by ProtoData.
+ */
+public interface ProtoDataApi {
```
armiol, line 32: I would consider getting rid of `Api` suffix, and replacing it with something closer to the Gradle domain. E.g. `ProtoDataExtension`.

That would make it difficult to mistreat the type and — for example — create a descendant unrelated to Gradle.

gradle-plugin/src/test/resources/empty-test/build.gradle.kts
```diff
@@ -37,7 +37,7 @@ buildscript {
 plugins {
     java
     id("com.google.protobuf")
-    id("io.spine.proto-data") version "0.1.8"
+    id("@PROTODATA_PLUGIN_ID@") version "@PROTODATA_VERIONS@"
```
armiol, line 40: There is a typo in `PROTODATA_VERIONS`.

It should probably be `PROTODATA_VERSION`. It's here and in all other places where it is used.

armiol (changes requested):
@alexander-yevsyukov LGTM in general. Please see my comments and suggestions.

dmdashenkov (approved):

gradle-api/src/main/kotlin/io/spine/protodata/gradle/Names.kt
alexander-yevsyukov: Will be migrated to newer version management API in a separate PR after it is available from the `tool-base` subproject. Hopefully soon.

armiol (approved):

PR #63, opened 2022-02-23
Republish attempt

dmdashenkov:
The previous build has failed because of an error while publishing.
The error has occurred because the script tried to publish the same artifact twice. This happens sometimes. Since there is going to be a shift in our publishing scripts towards variants, I suppose we could ignore this issue, as long as the publishing goes smooth this time.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #64, opened 2022-03-23
Better publishing

alexander-yevsyukov:
This PR adopts publishing improvements recently introduced in `config`. As the result of this, the following task (postponed in #61) was automatically resolved:
- [x] Do not publish `test` artifacts in `cli` module to public repositories.

Another improvement is adding the `protodata-` prefix for archives under the `protodata/lib/` in the distribution archive (named `protodata-setup.jar`) for archives representing our code (e.g. `cli-<version>.jar` or `compiler-<version>.jar`).

The latest `config` and `base` were also used.


armiol (approved):
LGTM.

PR #68, opened 2022-04-28
Manifested versions

alexander-yevsyukov:
This changes the way an artifacts version is stored and loaded in ProtoData, applying the mechanism recently [introduced in `tool-base`](https://github.com/SpineEventEngine/tool-base/pull/21).

There are also some improvements to the build scripts that were made because of issues experienced in builds under Windows 11:
 * `GitHubPackages.kt` was modified to extract the ZIP archive using the [zip4j](https://github.com/srikanth-lingala/zip4j) library, instead of using `unzip` command line utility which is not available in vanilla Windows.
 * `RunGradle.kt` was modified to use the `_out` directory instead of `build`. Using `build` for debug output of processes created by `ProcessBuilder` failed the `clean` command under Windows because of files already opened in the directory.

Changes to the build scripts will be applied to `config` upon completion of this PR.

### DIFF
--- a/buildSrc/src/main/kotlin/version-to-resources.gradle.kts
+++ b/buildSrc/src/main/kotlin/version-to-resources.gradle.kts
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2022, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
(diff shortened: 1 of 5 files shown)


alexander-yevsyukov:
The error `WindowsRegistry is not supported on this operating system.` which occurs in a Windows build is caused by this [issue in Gradle](https://github.com/gradle/native-platform/issues/274).

alexander-yevsyukov:
@armiol, @dmdashenkov, we still have the Gradle warning on `sourcesJar` does not explicitly or implicitly depending on `generateProto`. 

I've tried to quickly fix that by adding it explicitly similarly to the way it was recently introduced in `io.spine.internal.gradle.protobuf.ProtoTaskExtensions`, but failed: the warning persists. I guess it should be addressed at the level of `mc-java` which is applied.

armiol (approved):
LGTM.

As discussed vocally, we will also include the "old" variant of the versions file (`version.txt`) to be present JUST for this particular ProtoData version. That will help avoid us any potential migration issues. The `version.txt` file will be then dropped in the next update (i.e. `0.2.5` -> `0.2.6`).

@dmdashenkov FYI.

PR #69, opened 2022-07-12
Migrate to ProtoData-based `mc-java` and address some blocking issues

armiol:
This changeset updates the ProtoData to the latest `mc-java` artifacts, which are now based on ProtoData.

Additionally, a blocking issue was found and resolved. Previously, the custom Proto options defined as `repeated` were not handled properly by the ProtoData compiler:

```proto
extend google.protobuf.EnumValueOptions {

    // Possible file extensions associated with a file format.
    //
    // Only applicable to the `ConfigurationFormat` enum.
    //
    repeated string extension = 73981 [(internal) = true];
}

// The format of a custom configuration for ProtoData.
//
enum ConfigurationFormat {
    // ...

    PROTO_BINARY = 1 [(extension) = "bin", (extension) = "pb"];
  
     // ...
}
```

The codebase of ProtoData-under-development is now handled by the fresh `modelCompiler`, which in turn uses the previous version of ProtoData plugin. The issue caused the exception when processing the ProtoData's `configuration_proto`.

Therefore, to overcome the issue, this PR fixes it and advances the version to `0.2.5.1`:

* the parsing functionality is addressed,
* the part of the `main` code which contained the un-parseable code, was temporarily changed to get rid of the extension with the `repeated` field,
* the functionality covered with the corresponding testing.

It is an "intermediate" version placed between previously published `0.2.5` (used by `modelCompiler`, where issue is still present) and future `0.2.6` — which will be built on top of `modelCompiler` and ProtoData at `0.2.5.1`.

### DIFF
--- a/cli/src/test/kotlin/io/spine/protodata/cli/CliTest.kt
+++ b/cli/src/test/kotlin/io/spine/protodata/cli/CliTest.kt
@@ -40,7 +40,7 @@ import io.spine.protodata.cli.given.CustomOptionRenderer
 import io.spine.protodata.cli.given.TestOptionProvider
 import io.spine.protodata.cli.test.TestOptionsProto
 import io.spine.protodata.cli.test.TestProto
-import io.spine.protodata.tesst.Echo
+import io.spine.protodata.test.Echo
 import io.spine.protodata.test.ECHO_FILE
 import io.spine.protodata.test.EchoRenderer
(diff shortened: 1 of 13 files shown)


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #71, opened 2022-07-25
Use direct `Delivery`

armiol:
This changeset improves the performance of ProtoData-based apps (such as Validation) by switching to a direct `Delivery` mode in Spine's `ServerEnvironment`. 

Previously, the dispatching of signals were done by utilizing the `InboxStorage`, even though ProtoData-based apps generally do not operate in muti-threaded mode. It caused some excessive I/O operations (storing, reading and deleting data from `Inbox`es) and thus lower performance in terms of time-per-task.

This PR updates the `Delivery` settings to use a direct mode and thus eliminates some I/O and reduces the execution time. The tradeoff is that the library no longer prevents any concurrent modifications of entities — which should not happen at all if ProtoData-based apps operate in a single-threaded mode. 

Additionally, this PR explicitly switches off `modelCompiler`'s validation feature for two more Gradle modules (`codegen-java` and `cli`). Previously, for some reason, `validation-runtime` dependency appeared on both `compileClasspath` and `runtimeClasspath` of these modules, even though it was not used in any way. Also, it caused some version conflicts down the line, as `validation-runtime` dependency also appeared in the corresponding `pom.xml`s of the published artifacts.

The version of the ProtoData artifacts is set to `0.2.7`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -85,6 +85,7 @@ spinePublishing {
     artifactPrefix = "protodata-"
 }
 
+val coreVersion: String by extra
 allprojects {
     apply {
         from("$rootDir/version.gradle.kts")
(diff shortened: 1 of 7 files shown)


armiol:
@dmdashenkov PTAL.

dmdashenkov (approved):
LGTM.

PR #73, opened 2022-08-30
One run, many directories

dmdashenkov:
In this PR we allow ProtoData to process multiple directories with source code at once.

In CLI mode, this is achieved by supplying several paths for the `--src` and `--destination` arguments. The paths should be separated by the `:` symbol (or `;` on Windows). The number of source dirs must match the number of target dirs exactly.

In Gradle plugin mode, this is achieved by supplying several `subDirs`. Note that the plugin effectively limits the usage of this feature to processing several subdirectories within one root source directory. To process source files in several projects, one needs to run the ProtoData tasks for each project.

For compatibility reasons, in the Gradle extension, we are keeping the `subDir` property. In the Gradle task, we're keeping the `source` and `target` properties. These 3 properties are now deprecated and delegate to the new `subDirs`, `sources`, and `targets` respectively. The values of the old properties are treated as singular, i.e. there is no splitting and joining by the path separator; only one directory can be represented by the property.

When processing multiple source file sets, the `Renderer.render(..)` method is called multiple times. This is a change of behaviour compared to previous versions of ProtoData. However, there was never a guarantee of exactly one invocation of this method. Still, developers who create custom renderers should make sure files are not duplicated.

After all the changes and deprecations, the old configuration of ProtoData outputs the same behaviour are previously. However, the old properties [are to be removed](https://github.com/SpineEventEngine/ProtoData/issues/74) in future versions, so the users should migrate to the new API.


dmdashenkov:
@armiol, I'll see what I can do about those bot messages. Meanwhile, PTAL.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/LaunchProtoData.kt
```diff
@@ -72,13 +73,52 @@ public abstract class LaunchProtoData : JavaExec() {
      *
      * May not be available, if `protoc` built-ins were turned off, resulting in no source code
      * being generated. In such a mode `protoc` worked only generating descriptor set files.
+     *
+     * This property is deprecated. [sources] should be used in its stead. Accessing this property
```
armiol, line 77: Please fix the typo.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/LaunchProtoData.kt
```diff
@@ -72,13 +73,52 @@ public abstract class LaunchProtoData : JavaExec() {
...
+            sources = value.map { listOf(it) }
+        }
+        get() {
+            return sources.map { it.first() }
+        }
 
+    /**
+     * The path to the directory with the processed source code.
+     *
+     * This property is deprecated. [targets] should be used in its stead. Accessing this property
```
armiol, line 94: Same as above.

version.gradle.kts
```diff
@@ -34,4 +34,4 @@ val mcJavaVersion: String by extra("2.0.0-SNAPSHOT.94")
 val devProtoDataVersion: String by extra("0.2.5.1")
```
armiol, line 34: I think this can safely be set to `0.2.7`.

armiol (changes requested):

Also, please document the backward compatibility, at least in the PR description. As a matter of fact, I have just discovered that for inputs, we use just one of them (not having a "join"). It's fine too, but next time please arrange some means to tell about this design intention.

And have an issue to remove the previous inputs once the libraries migrate to the new API.

dmdashenkov:
@armiol, PTAL.

armiol (approved):
@dmdashenkov LGTM.

PR #75, opened 2022-09-26
Make `options.proto` and `time_options.proto` provided to ProtoData executable by default

armiol:
Previously, for all our current applications we had to supply ProtoData with some type, registering `options.proto` and `time_options.proto` programmatically. It was very inconvenient, and also lead to some unwanted dependencies onto artifacts containing this "registrar" type (in our case, it was a part of `mc-java`).

This changeset configures ProtoData to always know about `options.proto` and `time_options.proto`. It may become a problem in future if we decide to start actively changing the contents of Spine options. However, it was not the case for the past several years, and even in this case we will have to just update ProtoData.

Other than that, the GitHub Actions files were reverted, as they previously were overwritten by older versions.

Also, the Kotlin tests were configured, so that the console output contains the fact they actually `PASSED`. Previously, it was unclear whether they were launched at all.

One more change is an inclusion of `codegen-java` contents into the fat JAR. Without it, it is not possible to run ProtoData.

### DIFF
--- a/.github/workflows/build-on-ubuntu.yml
+++ b/.github/workflows/build-on-ubuntu.yml
@@ -7,16 +7,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
 
-      - uses: actions/setup-java@v2
+      - uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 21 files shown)


dmdashenkov (approved):
LGTM w/ a single formatting suggestion.

alexander-yevsyukov (approved):
LGTM

PR #76, opened 2022-09-27
Use only fat JARs in ProtoData's executable classpath

armiol:
Previously, ProtoData used a `protodata-cli` artifact as a Gradle dependency, when configuring the Gradle task on launching the ProtoData executable. However, this artifact has some dependencies of its own, which ultimately bring the whole Spine framework to the classpath. 

As we are moving towards running ProtoData in isolation, meaning no conflicts with Spine libraries-under-development, this artifact can no longer be used.

This changeset makes ProtoData to always configure a fat JAR (i.e. one without dependencies) as a sole dependency, when setting up a Gradle configuration for launching the executable. As a result, there is no dependencies onto Spine libs, and no conflicts in scope of Gradle builds.

Previously established `DevMode` (which allowed to switch to the fat JAR by passing a specific system property) is no longer needed, and has been removed within this PR.

This changeset will be accompanied by the corresponding changes to Validation library and `mc-java`.


armiol:
@alexander-yevsyukov @dmdashenkov PTAL.

dmdashenkov (approved):

PR #77, opened 2022-09-27
Ignore missing source dirs

dmdashenkov:
In this PR we add an ability for the user to specify source directory paths that do not lead to existing directories.

For that, in CLI mode, the `--ignore-missing` flag is introduced. If set, the CLI program ignores missing source directories, as long as at least one of them exists. If the source directory path leads to another FS object (i.e. a file or a symlink), just like before, an error occurs.

When launched from Gradle, the `--ignore-missing` behaviour is always enabled.


dmdashenkov:
@armiol, PTAL.

armiol (approved):
@dmdashenkov LGTM.

PR #78, opened 2022-09-28
Allow file presence checks

dmdashenkov:
Previously, it was cumbersome to check if a file is present in a `SourceFileSet` or not. Now we provide a simple API for that. The need for such API arises because previously, there was only one `SourceFileSet` per run and now there can be multiple.

### DIFF
--- a/compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
+++ b/compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
@@ -124,7 +124,7 @@ internal constructor(
     /**
      * Looks up a file by its path and throws an `IllegalArgumentException` if not found.
      *
-     * The [path] may be a relative or an absolute path the file.
+     * The [path] may be absolute or relative to the source root.
      */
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@armiol, PTAL.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
```diff
@@ -121,23 +122,31 @@ internal constructor(
...
      */
     public fun file(path: Path): SourceFile =
-        findFile(path).second
+        findFile(path).orElseThrow { IllegalArgumentException("File not found: `$path`.") }
 
-    private fun findFile(path: Path): Pair<Path, SourceFile> {
+    /**
+     * Looks up a file by its path.
+     *
+     * The [path] may be a relative or an absolute path the file.
```
armiol, line 135: It would be nice to describe "relative" to what exactly.

armiol (approved):
@dmdashenkov LGTM with a suggestion to consider.

PR #85, opened 2022-10-10
Bump versions, address deprecations

alexander-yevsyukov:
The general idea of this PR is making console output cleaner. This PR:
  * Bump dependencies on Spine artifacts.
  * Addresses Gradle warnings by adding explicit dependencies between tasks.
  * Rolls back the Protobuf dependency to version `3.19.6` to overcome the missing artefact issue under macOS running on M1.
  * Adds ErrorProne checks to the build.

### DIFF
--- a/buildSrc/src/main/kotlin/deps-between-tasks.kt
+++ b/buildSrc/src/main/kotlin/deps-between-tasks.kt
@@ -62,5 +62,6 @@ fun Project.configureTaskDependencies() {
         "sourcesJar".dependOn("createVersionFile")
         "dokkaHtml".dependOn("launchProtoDataMain")
         "dokkaJavadoc".dependOn("launchProtoDataMain")
+        ":publishPluginJar".dependOn("createVersionFile")
     }
 }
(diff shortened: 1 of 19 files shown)


buildSrc/src/main/kotlin/deps-between-tasks.kt
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * This extension is kept local to `configureTaskDependencies` extension
+     * to prevent its direct usage from outside.
+     */
+    fun String.dependOn(taskName: String) {
+        val whoDepends = this
+        val dependOntoTask: Task? = tasks.findByName(taskName)
+        dependOntoTask?.apply {
+            tasks.findByName(whoDepends)?.dependsOn(this)
+        }
```
dmdashenkov, line 55: Please use an `if` or a `let` instead. The meaning of `this` is not 100% clear from the first glance.

dmdashenkov: This is how we get JavaScript :)

dmdashenkov (approved):
LGTM with a suggestion.

buildSrc/src/main/kotlin/deps-between-tasks.kt
alexander-yevsyukov: Goot catch. I'll update `config` and `template` too then.

PR #87, opened 2022-10-15
Add `kotlin` `protoc` built-in when Java or Kotlin are enabled in a project

alexander-yevsyukov:
This PR teached ProtoData Gradle Plugin to turn `kotlin` `protoc` built-in when Java or Kotlin enabled in a project.

When working on this PR the issue of incompatibility with Android projects was [discovered](#88). We need to address it before the stable release.

Other notable changes:
  * Dependencies were simplified using the `Spine` object.
  * Integration tests were patched to allow running in a project which has both Java and Kotlin. We need a better way of doing this (#90).
  * properties `sourceRoot` and `targetRoot` were opened in `SourceFileSet`. Still, it did not allow to use these properties in the integration tests because tests seem to use the code embedded into `mc-java` rather than that defined in the build file.
  * `SourceFileSet` also got properties `size` and `isEmpty`. It was possible to obtain these values via `count` extension of `Iterable`, but doing it directly is a bit more efficient.


tests/protodata-extension/src/main/java/io/spine/protodata/test/uuid/UuidJavaRenderer.java
```diff
@@ -70,6 +70,9 @@ public final class UuidJavaRenderer extends JavaRenderer {
      */
     @Override
     protected void render(SourceFileSet sources) {
+        if (!sources.sourceRoot.endsWith("java")) {
```
alexander-yevsyukov, line 73: @armiol, this line breaks the build sayin that the `sourceRoot` property is `private`. It was indeed until I opened it.

It looks like indication that we test against the "dev" version, and not the one under the development. I've tried to force versions in the root build, but it didn't give much. I still get the "private" property error.

Let's discuss when time permits.


I've opened the

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Plugin.kt
```diff
@@ -174,6 +174,23 @@ private fun Project.configureWithProtobufPlugin(extension: Extension, version: S
     }
 }
 
+/**
+ * Verifies if the project has `java` plugin or `compileKotlin` or `compileTestKotlin` tasks.
+ *
+ * Even though current Protobuf support of Kotlin is based on Java codegen (and therefore
+ * it's likely that Java would be enabled in the project in the near future for Kotlin proto
```
armiol, line 181: I don't understand this bit about "near future". Could you please rephrase?

tests/protodata-extension/src/main/java/io/spine/protodata/test/annotation/AnnotationRenderer.java
```diff
@@ -45,6 +45,12 @@ public final class AnnotationRenderer extends JavaRenderer {
 
     @Override
     protected void render(SourceFileSet sources) {
+        //TODO:2022-10-18:alexander.yevsyukov: Use `sourceRoot` when it's public and
+        // remove `findFile()` check in the `renderFor()` method below.
+        //
```
armiol, line 50: Could you please refer to the GitHub issue here?

armiol (approved):

PR #93, opened 2022-10-19
Fix configuration of `compileKotlin`

alexander-yevsyukov:
This PR teaches ProtoData Plugin to configure `compileKotlin` tasks to eliminate duplicate declaration issue because of presence of the same code under `build/generated-proto/kotlin` and `generated/kotlin` directories.

Also, it changes the structure of tests, introducing `functionalTest` source set under `gradle-plugin` module.

The dependency on Jackson and its Databind were bumped to address security recommendations. Also JUnit dependencies were updated to latest versions.


dmdashenkov (approved):
LGTM.

armiol (approved):

PR #98, opened 2022-10-24
Fix custom source sets

### LINKED ISSUE #97: Unable to add a custom source set like `functionalTest`
Just tried to arrange a `testFixture` and `functionalTest` source sets in `mc-java-annotations` (following the detekt as an example). Got this:

```
Caused by: org.gradle.api.UnknownTaskException: Task with name 'launchProtoDataFunctionalTest' not found in project ':mc-java-annotation'.
        at org.gradle.api.internal.tasks.DefaultTaskCollection.createNotFoundException(DefaultTaskCollection.java:102)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.createNotFoundException(DefaultTaskCollection.java:46)
        at org.gradle.api.internal.DefaultNamedDomainObjectCollection.getByName(DefaultNamedDomainObjectCollection.java:333)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.getByName(DefaultTaskCollection.java:46)
        at io.spine.protodata.gradle.LaunchTask.get(LaunchTask.kt:50)
        at io.spine.protodata.gradle.plugin.PluginKt$configureProtobufPlugin$1$2.invoke(Plugin.kt:219)
        at io.spine.protodata.gradle.plugin.PluginKt$configureProtobufPlugin$1$2.invoke(Plugin.kt:206)
        at org.gradle.kotlin.dsl.KotlinClosure1.doCall(GroovyInteroperability.kt:94)
        at jdk.internal.reflect.GeneratedMethodAccessor62.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at org.gradle.util.internal.ClosureBackedAction.execute(ClosureBackedAction.java:73)
        at org.gradle.util.ConfigureUtil.configureTarget(ConfigureUtil.java:203)
...
```

If I try to add a dummy task like this:
```kotlin
val launchProtoDataFunctionalTest by tasks.registering {
}
```
I get `ConcurrentModificationException`:

```
A problem occurred configuring project ':mc-java-annotation'.
> java.util.ConcurrentModificationException (no error message)
```

alexander-yevsyukov:
This PR resolves #97 by delaying `LaunchProtoData` task creation to `afterEvaluate` to handle the cases of delayed creation of custom source sets.

Latest `config` was also applied.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/Jackson.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/Jackson.kt
@@ -26,7 +26,6 @@
 
 package io.spine.internal.dependency
 
-
 @Suppress("unused")
 object Jackson {
     const val version = "2.13.4"
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/Jackson.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/Jackson.kt
@@ -26,22 +26,31 @@
 
 package io.spine.internal.dependency
 
+
 @Suppress("unused")
 object Jackson {
     const val version = "2.13.4"
-    const val databindVersion = "2.13.4.2"
+    private const val databindVersion = "2.13.4.2"
+
+    private const val coreGroup = "com.fasterxml.jackson.core"
+    private const val dataformatGroup = "com.fasterxml.jackson.dataformat"
+    private const val moduleGroup = "com.fasterxml.jackson.module"
+
     // https://github.com/FasterXML/jackson-core
-    const val core = "com.fasterxml.jackson.core:jackson-core:${version}"
+    const val core = "$coreGroup:jackson-core:${version}"
     // https://github.com/FasterXML/jackson-databind
-    const val databind = "com.fasterxml.jackson.core:jackson-databind:${databindVersion}"
+    const val databind = "$coreGroup:jackson-databind:${databindVersion}"
+    // https://github.com/FasterXML/jackson-annotations
+    const val annotations = "$coreGroup:jackson-annotations:${version}"
+
     // https://github.com/FasterXML/jackson-dataformat-xml/releases
-    const val dataformatXml = "com.fasterxml.jackson.dataformat:jackson-dataformat-xml:${version}"
+    const val dataformatXml = "$dataformatGroup:jackson-dataformat-xml:${version}"
     // https://github.com/FasterXML/jackson-dataformats-text/releases
-    const val dataformatYaml = "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${version}"
+    const val dataformatYaml = "$dataformatGroup:jackson-dataformat-yaml:${version}"
+
     // https://github.com/FasterXML/jackson-module-kotlin/releases
-    const val moduleKotlin = "com.fasterxml.jackson.module:jackson-module-kotlin:${version}"
+    const val moduleKotlin = "$moduleGroup:jackson-module-kotlin:${version}"
+
     // https://github.com/FasterXML/jackson-bom
     const val bom = "com.fasterxml.jackson:jackson-bom:${version}"
-    // https://github.com/FasterXML/jackson-annotations
-    const val annotations = "com.fasterxml.jackson.core:jackson-annotations:${version}"
 }
--- a/gradle-plugin/src/functionalTest/resources/with-functional-test/build.gradle.kts
+++ b/gradle-plugin/src/functionalTest/resources/with-functional-test/build.gradle.kts
@@ -70,7 +70,8 @@ protobuf {
     }
 }
 
-@Suppress("UNUSED_VARIABLE") // `test` and `functionalTest`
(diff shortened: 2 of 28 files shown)


alexander-yevsyukov:
@armiol, @dmdashenkov, the error under Windows was caused by the call to `GradleProject.enableRunnerDebug()` I forgot to turn off. Somehow it causes resolution of Groovy dependency, which in turn fails because of [this issue](https://github.com/gradle/native-platform/issues/274).

Removing the `enableRunnerDebug()` call avoids the problem.

buildSrc/src/main/kotlin/io/spine/internal/dependency/Jackson.kt
```diff
@@ -26,22 +26,31 @@
 
 package io.spine.internal.dependency
 
+
```
armiol, line 29: This line looks weird. Do we need it?

gradle-plugin/src/functionalTest/resources/with-functional-test/build.gradle.kts
```diff
@@ -0,0 +1,82 @@
...
+protobuf {
+    protoc {
+        artifact = io.spine.internal.dependency.Protobuf.compiler
+    }
+}
+
+@Suppress("UNUSED_VARIABLE") // `test` and `functionalTest`
+testing {
+    suites {
+        val test by getting(JvmTestSuite::class) {
```
armiol, line 76: Please comment on this approach. The unused variables still raise questions, even though the warning is suppressed.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Plugin.kt
```diff
@@ -216,12 +248,37 @@ private fun Project.configureProtobufPlugin(extension: Extension, version: Strin
...
-                launchTask.dependsOn(task)
+                handleLaunchTaskDependency(extension, sourceSet, task)
             }
         }
         generatedFilesBaseDir = "$buildDir/generated-proto/"
     }
+}
+
+/**
+ * Makes a [LaunchProtoData], if it exists for the given [sourceSet]
```
armiol, line 259: The comma is missing at the end of this line.

version.gradle.kts
```diff
@@ -24,11 +24,11 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-val baseVersion: String by extra("2.0.0-SNAPSHOT.114")
-val coreVersion: String by extra("2.0.0-SNAPSHOT.116")
+val baseVersion: String by extra("2.0.0-SNAPSHOT.115")
```
armiol, line 27: Looking at evolving `Spine` object, I wonder what's the point of maintaining both these versions of Spine libraries along with those defined by `Spine`. Such an approach is very much prone to errors, considering how often our libs are updated. 

Previously, I have encountered some issues caused by using different sources for the versions, in the scope of a single repository. I.e., in one use-case, the version from `versions.gradle.kts` was used, and in another use-case it was grabbed from `Spine` object. And they were different. Took me a couple of hours to figure out, back then.

Please consider addressing this discrepancy.

version.gradle.kts
```diff
@@ -24,11 +24,11 @@
...
+val coreVersion: String by extra("2.0.0-SNAPSHOT.117")
 val toolBaseVersion: String by extra("2.0.0-SNAPSHOT.111")
 val mcVersion: String by extra("2.0.0-SNAPSHOT.90")
-val mcJavaVersion: String by extra("2.0.0-SNAPSHOT.102")
+val mcJavaVersion: String by extra("2.0.0-SNAPSHOT.103")
 
 // The version of ProtoData being developed.
```
armiol, line 33: I think it's more like the version to publish. We had this comment about "being developed" to distinguish this version from the one used to build these sources (`protoDataDevVersion`), which is no longer needed, as it is now embedded into `modelCompiler`, as far as I understand.

armiol (approved):
LGTM with some comments to consider.

version.gradle.kts
alexander-yevsyukov: I think, we'd better get rid of all those constants in the `version` file in favor of using default parameters of the Kotlin `Spine` dependency object.

The implementation of `Spine` is currently tricky. Not to address in this PR, but we need it much simpler than now.

version.gradle.kts
alexander-yevsyukov: Added issue: https://github.com/SpineEventEngine/config/issues/418

version.gradle.kts
alexander-yevsyukov: Added issue: https://github.com/SpineEventEngine/config/issues/418.

alexander-yevsyukov: I've removed constants from `version.gradle.kts` in this PR. But we need more work on this. The added issue is about it.

PR #105, opened 2022-11-04
Inline insertion points

dmdashenkov:
### Inline insertion points

We enable ProtoData renderer developers to add insertion points inside existing lines of code. This will be useful for tweaking generated API.

When locating where to put such insertion points, users specify the line and column number for where in the existing code the insertion must take place. ProtoData will add a comment in that place. It's worth noting that the inline insertion points should be placed with care, so that not to break compilability of the existing code.

The new insertion points would look like this:
 - Whole-line points are not changed much, except that `Language` implementations should create block comments now instead of single-line comments. Here is a Java example:
```diff

  package com.acme.example;
- // INSERT:afterPackage:com.acme.example.Hello
+ /* INSERT:afterPackage:com.acme.example.Hello */

```
 - Inline insertion points are placed exactly at the column indicated by the SPI user. Here is a Java example:
```java

public com.acme.example. /* INSERT:returnTypeAnnotation:build */ Hello build() {

```
When code is inserted at this point, it will appear after the insertion point. `Renderer` authors can rely on one blank space between the inserted code and the preexisting code. 
Suppose, we would to insert `@javax.annotation.Nullable` at this point. The resulting code would look like this:
```java

public com.acme.example. /* INSERT:returnTypeAnnotation:build */@javax.annotation.Nullable Hello build() {

```

### Repeating insertion points

Also, in this PR, we allow insertion points to be repeated many times in a single file.

This means that the same insertion point can appear in a file many times. The code added at that insertion point will also be exactly the same each time.

Inline insertion points can repeat in multiple lines in a file and multiple times in a single line.

For this, the new version of `InsertionPoint.locate(..)` returns not one, but a set many of text coordinates.
Also, to simplify the non-repeating insertion point case, a helper `NonRepeatingInsertionPoint` is introduced. It simply wraps the desired text coordinates into a singleton set.


dmdashenkov:

There are a lot of files, but __most__ of them are just `config` changes.

armiol:
@dmdashenkov could you please append the PR description with some samples of insertion points? It would be helpful to understand the restrictions, and whether the "old" whole-line insertion points look differently (or not).

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -26,13 +26,28 @@
...
+import io.spine.text.TextFactory.text
 
 /**
  * A point is a source file, where more code may be inserted.
  */
-public interface InsertionPoint {
+public interface InsertionPoint : CoordinatesFactory {
+
+    public companion object {
+        public const val COMMENT_PADDING_LENGTH: Int = 8
```
armiol, line 49: Please document.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
+     * @param text the existing code
+     * @return the coordinates in the file where the insertion point should be added
+     * @see SourceFile.at
+     * @see SourceFile.atInline
+     */
+    public fun locate(text: Text): FileCoordinates
+}
+
+/**
+ * A factory of [FileCoordinates] instances.
```
armiol, line 100: Maybe, it could be a Kotlin extension to `FileCoordinates.Builder`?

E.g.
```kotlin
FileCoordinates.newBuilder().at(42, 0)
```
Then, users won't have to remember that there is one more type (i.e. `CoordinatesFactory`) producing the instances.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
+            .build()
+
+    /**
+     * Creates coordinates pointing at the first line in the file.
+     */
+    public fun startOfFile(): FileCoordinates =
+        atLine(0)
+
+    /**
+     * Creates coordinates pointing at the last line in the file.
```
armiol, line 131: Is this after the last symbol of the file?

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
+        FileCoordinates.newBuilder()
+            .setInline(
+                Position.newBuilder()
+                    .setLine(line)
+                    .setColumn(column)
+            )
+            .build()
+
+    /**
+     * Creates coordinates pointing at a specific line in the file.
```
armiol, line 117: Is this "at the beginning of the line"?

armiol (changes requested):
LGTM in general.

Please see my comments. In particular, it would be nice to be a bit more specific in some of documentation pieces.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -26,13 +26,28 @@
 
 package io.spine.protodata.renderer
 
+import com.google.common.flogger.StackSize.FULL
+import com.google.protobuf.Empty
+import io.spine.logging.Logging.loggerFor
+import io.spine.protodata.FileCoordinates
```
alexander-yevsyukov, line 32: Why is it `FileCoordinates`, but not `TextCoordinates` or `SourceCodeCoordinates`?

A file could be binary, what if it's not a file at all? What if it's a piece of printed text, and we help Elon to sort out Twitter developers?.. :) Would these things apply to a paper roll?

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -26,13 +26,28 @@
...
+import io.spine.protodata.FileCoordinates.SpecCase.END_OF_FILE
+import io.spine.protodata.FileCoordinates.SpecCase.INLINE
```
alexander-yevsyukov, line 34: > `SpecCase`

Is it “special case” or “specification case” or something else?

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
      * @see LineNumber
      */
-    public fun locate(lines: List<String>): LineNumber
+    @Deprecated("Use locate(Text) instead.")
+    public fun locate(lines: List<String>): LineNumber {
+        val coords = locate(text(lines))
+        return when (coords.specCase) {
+            WHOLE_LINE -> LineNumber.at(coords.wholeLine)
+            INLINE -> {
+                loggerFor(InsertionPoint::class.java)
```
alexander-yevsyukov, line 73: Please extract it into a method. This is, obviously, an important piece of code because we have highly visible  logging in this branch. So, a method, which should be documented is highly desirable.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
+                    .atWarning()
+                    .withStackTrace(FULL)
+                    .log("`locate(List<String>)` does not support inline insertion. " +
+                            "A whole line insertion will be generated.")
+                LineNumber.at(coords.inline.line)
+            }
+            END_OF_FILE -> LineNumber.endOfFile()
+            NOT_IN_FILE -> LineNumber.notInFile()
```
alexander-yevsyukov, line 81: How can we have a “line number not in file”? It's just “not in file”. It can be anywhere around the file. Lines are just not applicable in this case.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
+            else -> throw IllegalStateException("Unexpected file coordinates `$coords`.")
```
alexander-yevsyukov, line 82: detekt hints to use `error()` instead of `IllegalStateException`. I tend to agree. It's easier to read.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +64,84 @@ public interface InsertionPoint {
...
+     * Locates the site where the insertion point should be added.
+     *
+     * An insertion point should only appear once in a file.
+     *
+     * @param text the existing code
+     * @return the coordinates in the file where the insertion point should be added
+     * @see SourceFile.at
+     * @see SourceFile.atInline
+     */
+    public fun locate(text: Text): FileCoordinates
```
alexander-yevsyukov, line 96: It does not feel right. We locate text and get file coordinates. Come on!.. :)

File coordinates feels like drive/directory/.../directory thing.

alexander-yevsyukov (changes requested):
Please see my comments. If we need to talk, let's gather and have a quick chat.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
dmdashenkov: `CoordinatesFactory` is a trait added to `InsertionPoint`s. It's not meant to be instantiated by the user themselves. I've added the description to reflect that in the doc.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
dmdashenkov: `LineNumber` is the old type that is being deprecated and replaced by `TextCoordinates`, so I won't touch it now.

armiol:
@dmdashenkov just in case you missed my previous comment a few days ago: 

...could you please append the PR description with some samples of insertion points? It would be helpful to understand the restrictions, and whether the "old" whole-line insertion points look differently (or not).

compiler/src/main/proto/spine/protodata/insertion.proto
```diff
@@ -0,0 +1,60 @@
...
+// A pointer to a place in a text file.
+message TextCoordinates {
+
+    oneof kind {
+        option (is_required) = true;
+
+        // Points at specific line and column.
+        spine.text.Position inline = 1 [(validate) = true];
+
+        // Points at a whole line.
```
armiol, line 51: Is this a pointer at the beginning of the line, or at the end of the line?

compiler/src/main/proto/spine/protodata/insertion.proto
```diff
@@ -0,0 +1,60 @@
...
+        uint32 whole_line = 2;
+
+        // Points at the last line of the file.
```
armiol, line 54: Does it refer to the point after the last symbol of the last line, or to the point before the first symbol of the last line?

armiol (changes requested):
@dmdashenkov 
Please see my comments to the PR (really confuse me :(), and my standalone comment right above.

compiler/src/test/kotlin/io/spine/protodata/InsertionPointTest.kt
```diff
@@ -0,0 +1,117 @@
...
+                     const val I_AM_CONSTANT: String = "!!"
+                 }
+                
+                 fun letsHaveFun(): String {
+                     return "trololo"
+                 }
+            }
+            """.trimIndent())
+            pipeline = Pipeline(
+                listOf(),
```
alexander-yevsyukov, line 77: Can we have parameter names here, please?

compiler/src/test/kotlin/io/spine/protodata/PipelineTest.kt
```diff
@@ -155,16 +159,34 @@ class `'Pipeline' should` {
...
+    fun `write into inline insertion points`() {
+        val sourceFile = write("ClassWithMethod.java", """
+            class ClassWithMethod {
+                public java.lang.String foo() {
+                    return "happy halloween";
+                }
+            }
+        """.trimIndent())
+        Pipeline(
+            listOf(),
```
alexander-yevsyukov, line 183: Please have parameter names here too.

compiler/src/test/kotlin/io/spine/protodata/InsertionPointTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        @Test
+        fun `a specific line and column`() {
+            pipeline()
+            val file = path / FILE_NAME
+            val contents = file.readLines()
+            assertThat(contents)
+                .isNotEmpty()
+            assertThat(contents[3])
+                .contains("I_AM_CONSTANT:/* ${LINE_FOUR_COL_THIRTY_THREE.codeLine} */      String")
```
alexander-yevsyukov, line 114: The insertion point name is a bit puzzling. What does it mean?

alexander-yevsyukov: Why do we have extra space before `String`?

alexander-yevsyukov (changes requested):
Please see my comments and questions.

compiler/src/test/kotlin/io/spine/protodata/InsertionPointTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import kotlin.io.path.createFile
+import kotlin.io.path.div
+import kotlin.io.path.readLines
+import kotlin.io.path.writeText
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+
+class `Insertion points should` {
```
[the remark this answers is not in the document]
dmdashenkov: I thought we agreed NOT to use `@DisplayName`?

compiler/src/test/kotlin/io/spine/protodata/InsertionPointTest.kt
dmdashenkov: The extra space is there to account for all possible comment-closing syntax. Since when we insert code at the insertion points we don't know how the comment is formatted, we make sure to leave enough chars to wrap the insertion point label into a comment. The number of chars we pad is hard-coded to be 8.

This is the trick I was talking about in our call some time ago. An alternative would be to incorporate the column number into the insertion point text somehow, but that is also not very easy, since, as you pointed out, we would have to shift any other column numbers when inserting new code into a line.

The name of this test-only insertion point does not have a hidden meaning and reflects the place in the text where it's located: line 4, column 33.

armiol:
@dmdashenkov can we rearrange the whitespace, so that all of excess (reserved?) space symbols reside inside the comment, not outside of it?

The idea is that after the replacement, the resulting code does not have weird spacing.

compiler/src/test/kotlin/io/spine/protodata/InsertionPointTest.kt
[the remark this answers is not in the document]
alexander-yevsyukov: If we use `DisplayName` we:

1. Avoid nasty code scanning warnings from `github-code-scanning` bot, complaining like “If a source code contains...”.
2. We could refer to a test suite in KDoc and our communications via a test suite class name.
3. Could use backticks and other symbols however we like.

This concerts top level classes. Inner oncs can and should be mostly backticked.

dmdashenkov:
@armiol, there is no _replacing_ going on, only _inserting_. The comment never goes away, so that it can be used by multiple renderers.

compiler/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
alexander-yevsyukov: @dmdashenkov, `FileCoordinates` and `SpecCase` naming issues are better to resolve in this PR.

If we need to gather and discuss it, let's do it.

armiol:
@dmdashenkov even so:

a) we could strip the comments automatically in the future, as a next step _after_ the codegen;
and
b) with less "code" space symbols, we will get less noise from static code analyzers, if/when we decide to analyze our code with them.

Consider using ProtoData for the code generation in some commercial product. In this case, we will definitely consider the removal of all the "technical" comments, before shipping the generated code as, say, a part of public API.

alexander-yevsyukov (changes requested):
We're close, but still have some naming issues to address. Please see my comments.

alexander-yevsyukov:
@dmdashenkov, [this PR](https://github.com/SpineEventEngine/ProtoData/pull/107) introduced some changes that affect this one. Please look into #107, it may simplify resolving conflicts a bit.

compiler/src/main/proto/spine/protodata/insertion.proto
```diff
@@ -0,0 +1,60 @@
...
+    oneof kind {
+        option (is_required) = true;
+
+        // Points at specific line and column.
+        spine.text.Position inline = 1 [(validate) = true];
+
+        // Points at the start of a specific line.
+        uint32 whole_line = 2;
+
+        // Points at the start of the last line of the file.
```
armiol, line 54: I know I have been asking that before, but wouldn't it be more intuitive to point to the end of the last line of the file?

It's an "end of file", after all.

compiler/src/main/proto/spine/protodata/insertion.proto
```diff
@@ -0,0 +1,60 @@
...
+        google.protobuf.Empty end_of_file = 3;
```
armiol, line 55: Another thing that just came up to me. `Empty` has only a default instance, right? If so, how this `oneof` would distinguish this `end_of_file` being "enabled" from the one being "disabled"?

armiol (changes requested):
@dmdashenkov please see two more questions inline, in addition to my comment on placing the whitespace.

alexander-yevsyukov:
> white-space gap to ensure no compilation errors are accidentally created

@dmdashenkov, @armiol, how having a gap would save us from user errors? The on-off problem is vast and eternal, as many other counting problems on programming. But do we really think that 8 spaces can cure it once and for all?

compiler/src/test/kotlin/io/spine/protodata/InsertionPointTest.kt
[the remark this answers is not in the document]
alexander-yevsyukov: Found the 4th reason:
  * Backticked test suite names causes difficulties for IDEA when running all the tests from the top level class declaration. It complains it couldn't find the tests in `*Spec` or `*Test`.

compiler/src/main/proto/spine/protodata/insertion.proto
dmdashenkov: The `oneof` knows that a field is set, even if the value is default. It works with any type and value, including zeros for numbers etc.

dmdashenkov:
In order to move the spaces after the insertion point comment into the comment, one should simply alter the way comments are generated (see `Language` and its subtypes). There, add the spaces after the insertion point tag and before the closing `*/`. I would recommend against reducing the number of padding chars that a renderer expects, as we don't know what we'll have to fit into them in the future.

.github/workflows/build-on-ubuntu.yml
```diff
@@ -22,23 +22,13 @@ jobs:
...
-      - name: Publish Unit Test Report
+      - name: Publish Test Report
         uses: mikepenz/action-junit-report@v3.5.2
         if: always() # always run even if the previous step fails
         with:
-          check_name: Regular Unit Tests
           report_paths: '**/build/test-results/test/TEST-*.xml'
           require_tests: true # will fail workflow if test reports not found
 
-      # See: https://github.com/marketplace/actions/junit-report-action
```
alexander-yevsyukov, line 33: I think we need these changes restored. We'll need those reports if something goes wrong in GitHub build.

dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL at the view-based approach for inline insertion points.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -202,16 +239,24 @@ private constructor(
...
-        return code
+        return text().value
     }
 
     /**
      * Obtains the entire content of this file as a list of lines.
      */
     public fun lines(): List<String> {
-        return lineSplitter.splitToList(code())
+        return text().splitLines()
```
alexander-yevsyukov, line 251: Why don't we have `lines` as a property of `Text`, which returns a list of `Line`s? And a `Line` can have a `number` property and all necessary operations we already have and may want to have in the very near future.

I'm not suggesting to do it right in this PR. But we definitely need more powerful abstractions.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -283,9 +328,110 @@ internal constructor(
...
+}
+
+/**
+ * A [SourceAtPointInline] at a point that could not be located.
+ *
+ * No code is generated by this implementation.
+ */
+internal object Unlocated : SourceAtPointInline, Logging {
+    override fun add(codeFragment: String) {
+        // Do nothind.
```
alexander-yevsyukov, line 354: If you do it like this, detekt won't complain:

```kotlin
override fun add(codeFragment: String) = Unit
```

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -283,9 +328,110 @@ internal constructor(
...
+        }
+    }
+}
+
 private fun Iterable<String>.linesToCode(indentLevel: Int): String =
     joinToString(lineSeparator()) {
         INDENT.repeat(indentLevel) + it
     }
 
 private const val INDENT: String = "    "
```
alexander-yevsyukov, line 415: This should probably depend on language or source file settings.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -283,9 +328,110 @@ internal constructor(
...
+
+/**
+ * A numbered line of code in a file.
+ */
+private data class CodeLine(val lineIndex: Int, val content: String) {
```
alexander-yevsyukov, line 420: Ha, see, we do need numbers together with lines!  Why isn't it a `Line` in a Text library then?

test-env/src/main/kotlin/io/spine/protodata/test/AnnotationInsertionPointPrinter.kt
```diff
@@ -0,0 +1,75 @@
...
+import io.spine.protodata.TextCoordinates
+import io.spine.protodata.renderer.InsertionPoint
+import io.spine.protodata.renderer.InsertionPointPrinter
+import io.spine.protodata.splitLines
+import io.spine.text.Text
+import io.spine.tools.code.CommonLanguages.Java
+import kotlin.text.RegexOption.DOT_MATCHES_ALL
+
+
+public class AnnotationInsertionPointPrinter: InsertionPointPrinter(Java) {
```
alexander-yevsyukov, line 38: Please document this class. I understand its under tests, but we need to treat this code equally as the production code.

test-env/src/main/kotlin/io/spine/protodata/test/AnnotationInsertionPointPrinter.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    override fun supportedInsertionPoints(): Set<InsertionPoint> =
+        AnnotationInsertionPoint.values().toSet()
+}
+
+public enum class AnnotationInsertionPoint : InsertionPoint {
```
alexander-yevsyukov, line 44: Please document this enum class.

test-env/src/main/kotlin/io/spine/protodata/test/NullableAnnotationRenderer.kt
```diff
@@ -0,0 +1,45 @@
...
+package io.spine.protodata.test
+
+import io.spine.protodata.renderer.Renderer
+import io.spine.protodata.renderer.SourceFileSet
+import io.spine.protodata.test.AnnotationInsertionPoint.BEFORE_RETURN_TYPE_METHOD_FOO
+import io.spine.protodata.test.AnnotationInsertionPoint.IMPORT
+import io.spine.tools.code.CommonLanguages.Java
+import kotlin.io.path.Path
+
+public class NullableAnnotationRenderer : Renderer(Java) {
```
alexander-yevsyukov, line 36: Please document this class.

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +81,93 @@ public interface InsertionPoint {
...
+            .setEndOfFile(Empty.getDefaultInstance())
+            .build()
+
+    /**
+     * Creates coordinates that do not point at anywhere in the text.
+     */
+    public fun nowhere(): TextCoordinates =
+        TextCoordinates.newBuilder()
+            .setNotInFile(Empty.getDefaultInstance())
+            .build()
```
alexander-yevsyukov, line 170: Isn't this a constant?

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +81,93 @@ public interface InsertionPoint {
...
+    public fun startOfFile(): TextCoordinates =
+        atLine(0)
+
+    /**
+     * Creates coordinates pointing at the point after the last line in the text.
+     */
+    public fun endOfFile(): TextCoordinates =
+        TextCoordinates.newBuilder()
+            .setEndOfFile(Empty.getDefaultInstance())
+            .build()
```
alexander-yevsyukov, line 162: Isn't this a constant?

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +81,93 @@ public interface InsertionPoint {
...
+
+    /**
+     * Creates coordinates that do not point at anywhere in the text.
```
alexander-yevsyukov, line 165: This is a bit controversial comment for a thing called `InsertionPoint`. Where are we going to insert, if at all then? Please elaborate.

alexander-yevsyukov (changes requested):
Please see my comments.

alexander-yevsyukov (changes requested):
Please see my comments.

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
```diff
@@ -49,7 +81,93 @@ public interface InsertionPoint {
...
+            .withStackTrace(FULL)
+            .log(
+                "`locate(List<String>)` does not support inline insertion. " +
+                        "A whole line insertion will be generated."
+            )
+
+    /**
+     * Locates the site where the insertion point should be added.
+     *
+     * An insertion point should only appear once in a file.
```
armiol, line 111: Taking into account our last conversation, this sentence is not too clear. Can we create an insertion point for, say, a repeating fragment of code picking it via a regular expression? Such an insertion point may serve to suppress common warnings.

In any case, I think this feature requires better documentation.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -110,10 +119,38 @@ private constructor(
...
+     * @see atInline
      */
     public fun at(insertionPoint: InsertionPoint): SourceAtPoint =
         SourceAtPoint(this, insertionPoint)
 
+    /**
+     * Creates a new fluent builder for adding code at the given inline [insertionPoint].
+     *
+     * If the [insertionPoint] is not found in the code, no action will be performed as the result.
+     * If there are more than one instances of the same insertion point, the code will be added to
```
armiol, line 132: "instance"

And again, here we speak about multiple instances of the same insertion point. See my comment on this matter above.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
armiol, line 354: "nothing"

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
armiol, line 420: It looks at least a descendant of `Line`, which probably exists somewhere in `text` library.

armiol (changes requested):
@dmdashenkov please see my comments. Perhaps, we should discuss them vocally at some point.

alexander-yevsyukov:
Here's the suggested code for a `companion object` somewhere near `TextCoordinates` declareation:

```kotlin
    public companion object {

        public fun at(line: Int, column: Int): TextCoordinates {
            return textCoordinates {
                inline = position {
                    cursor = cursor {
                        this.line = line
                        this.column = column
                    }
                }
            }
        }

        public fun atLine(line: Int): TextCoordinates = textCoordinates { wholeLine = line }
        
        private val empty: Empty = Empty.getDefaultInstance()
        public val startOfFile: TextCoordinates = atLine(0)
        public val endOfFile: TextCoordinates = textCoordinates { endOfFile = empty }
        public val nowhere: TextCoordinates = textCoordinates { notInFile = empty }
    }
```

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -110,10 +119,38 @@ private constructor(
...
+     * the same insertion point may be used in both modes.
+     *
+     * @see at
+     */
+    public fun atInline(insertionPoint: InsertionPoint): SourceAtPointInline {
+        val points = sources.querying.select<InsertedPoints>()
+            .find(filePath { value = relativePath.toString() })
+            .orElseThrow()
+        val point = points.pointList.filter { it.label == insertionPoint.label }.firstOrNull()
+        return if (point != null) {
```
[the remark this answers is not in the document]
alexander-yevsyukov: @dmdashenkov,  how about having it like this:
```kotlin
return point?.let { Located(this@SourceFile, point) } 
            ?: Unlocated
```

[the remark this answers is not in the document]
alexander-yevsyukov: @armiol, or, probably, we need to relax the rule of always having curly braces, making the exclusion for quasi-ternary `if/else`. + @dmdashenkov 

What's your take?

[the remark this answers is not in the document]
alexander-yevsyukov: I still think that `?.let {} ?:` is more idiomatic to Kotlin, because it deals with nullability. And this is what `?.` and `?:` are about.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -202,16 +239,24 @@ private constructor(
...
      */
     public fun lines(): List<String> {
-        return lineSplitter.splitToList(code())
+        return text().splitLines()
+    }
+
+    /**
+     * Obtains the text content of this source file.
+     */
+    public fun text(): Text {
```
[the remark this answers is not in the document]
dmdashenkov: This is a mutable property. We cannot use `lazy` with something that can change.
Additionally, I'd rather have `text()` than `getText()` in Java.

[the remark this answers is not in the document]
dmdashenkov: Also, since this may or may not force an IO operation, this should probably be a `fun` in Kotlin as well.

dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL. The issue of forming a `Line` into a real type is left unresolved for now. I suggest we discuss it an, if needed, do another PR.

On the `let` vs `if` question, I suggest we form a meaningful guideline and get back to it later as well. For now, I stay unconvinced. Here is a [link](https://kotlinlang.org/docs/null-safety.html#checking-for-null-in-conditions) to Kotlin's dot that uses `if`.

armiol:
@dmdashenkov please use "re-request review" feature next time, as I am drowning in PRs sometimes, and they may slip away.

dmdashenkov:
@armiol, will do.

api/src/main/kotlin/io/spine/protodata/renderer/SourceAtLine.kt
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @see SourceFile.at
+ */
+public class SourceAtLine
+internal constructor(
+    private val file: SourceFile,
+    private val point: InsertionPoint
+) {
+
+    private var indentLevel: Int = 0
```
armiol, line 42: Just in case, we also have `io.spine.string.Indent`, and we can use it here.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
```diff
@@ -259,6 +262,13 @@ internal constructor(
         }
     }
 
+    /**
+     * Initialize this set with a [Querying] instance for performing internal checks.
```
armiol, line 266: "Initializes"

api/src/main/proto/spine/protodata/insertion.proto
```diff
@@ -0,0 +1,68 @@
...
+    FilePath file = 1;
+
+    // The label of the insertion point.
+    string label = 2 [(required) = true, (column) = true];
+
+    // The code representation of the insertion point.
+    //
+    // Typically, this is a comment containing the insertion point label.
+    //
+    string representation_in_code = 3 [(required) = true];
```
armiol, line 56: They are probably `distinct` as well.

armiol (changes requested):

One thing that I don't _quite_ understand is that example of an inline insertion point in the PR description. Maybe, we could make it compilable? At the moment my internal compiler breaks on it, but maybe it's just me.

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPointPrinter.kt
```diff
@@ -53,30 +60,140 @@ public abstract class InsertionPointPrinter(
...
+        if (!coords.compatibleWith(definingType)) {
+            throw RenderingException(
+                "One insertion point (${insertionPoint::class.qualifiedName}) cannot be " +
+                        "whole-line and inline at the same time."
+            )
+        }
+    }
+}
+
+private fun TextCoordinates.compatibleWith(type: TextCoordinates.KindCase) = when (kindCase) {
```
alexander-yevsyukov, line 195: Please document this.

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPointPrinter.kt
```diff
@@ -53,30 +60,140 @@ public abstract class InsertionPointPrinter(
...
+        when (coords.kindCase) {
+            INLINE -> return INLINE
+            WHOLE_LINE, END_OF_TEXT -> return WHOLE_LINE
+            else -> {} // Keep searching for the type.
         }
     }
+    return null
+}
+
+private fun Iterable<TextCoordinates>.ensureSameType(insertionPoint: InsertionPoint,
```
alexander-yevsyukov, line 183: Please document this.

api/src/main/kotlin/io/spine/protodata/renderer/InsertionPointPrinter.kt
```diff
@@ -53,30 +60,140 @@ public abstract class InsertionPointPrinter(
...
+    if (position < 0 || position >= length) {
+        throw RenderingException(
+            "Line does not have column $position: `$this`."
+        )
+    }
+}
+
+@Suppress("ReturnCount")
+    // As this function is pretty small, it's easy to read even with 3 return statements.
+private fun Iterable<TextCoordinates>.precedentType(): TextCoordinates.KindCase? {
```
alexander-yevsyukov, line 172: Please document this.

api/src/main/kotlin/io/spine/protodata/renderer/SourceAtLine.kt
```diff
@@ -0,0 +1,101 @@
...
+     *      code lines
+     */
+    public fun add(vararg lines: String): Unit =
+        add(lines.toList())
+
+    /**
+     * Adds the given code lines at the associated insertion point.
+     *
+     * @param lines
+     *      code lines
```
alexander-yevsyukov, line 68: Please align the parameter description also adding trailing period.

api/src/main/kotlin/io/spine/protodata/renderer/SourceAtLine.kt
```diff
@@ -0,0 +1,101 @@
...
+    public fun withExtraIndentation(level: Int): SourceAtLine {
+        Preconditions.checkArgument(level >= 0, "Indentation level cannot be negative.")
+        indentLevel = level
+        return this
+    }
+
+    /**
+     * Adds the given code lines at the associated insertion point.
+     *
+     * @param lines
```
alexander-yevsyukov, line 58: Please align the parameter description also adding trailing period.

api/src/main/kotlin/io/spine/protodata/renderer/SourceAtLine.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+private fun Iterable<String>.linesToCode(indentLevel: Int): String =
+    joinToString(System.lineSeparator()) {
+        INDENT.repeat(indentLevel) + it
+    }
+
+/**
+ * A four-spaces indent.
```
alexander-yevsyukov, line 94: Not in this PR, but we'll need to make it more flexible than now.

alexander-yevsyukov (changes requested):
Please see my comments.

dmdashenkov:

The example in the PR header is compilable, but, as it is not a common case of use of annotations, GitHub's markup system fails to highlight it properly, I'm guessing.

armiol (approved):
@dmdashenkov LGTM with a minor comment to address.

alexander-yevsyukov (approved):
LGTM, with one minor request to address before the merge.

PR #107, opened 2022-11-09
Put events into subpackages

alexander-yevsyukov:
This PR:
  * Enforces [non-parallel builds with Gradle](https://stackoverflow.com/questions/38298695/gradle-disable-all-incremental-compilation-and-parallel-builds).
  * Improves configuring version used for Java and Kotlin compilation.
  * Moves generated event types under `event` subpackages.
  * Migrates to `OptionsProvider` from `base`.

The mystery of not being able Kotlin DSL introduced under the `compiler` modules is still not solved. 
And more attempts would be made in the following PRs.


build.gradle.kts
```diff
@@ -104,100 +104,39 @@ allprojects {
...
-        useJUnitPlatform()
-
-        testLogging {
-            events = setOf(PASSED, FAILED, SKIPPED)
-            showExceptions = true
-            showCauses = true
-        }
-    }
-
-    val javaVersion = JavaVersion.VERSION_11.toString()
```
armiol, line 157: I would still use this expression inside `BuildSettings`. That seems less error-prone. 

Or... (con'd below)

build.gradle.kts
```diff
@@ -104,100 +104,39 @@ allprojects {
...
-    tasks.withType<KotlinCompile> {
-        setFreeCompilerArgs()
-    }
+subprojects {
+    applyPlugins()
+    setDependencies()
+    forceConfigurations()
 
-    val dokkaJavadoc by tasks.getting(DokkaTask::class)
+    val javaVersion = JavaLanguageVersion.of(BuildSettings.javaVersion)
```
armiol, line 132: ...or, even the instance of `JavaLanguageVersion`.

compiler/src/test/kotlin/io/spine/protodata/CompilerEventsSpec.kt
```diff
@@ -33,84 +33,104 @@ import com.google.protobuf.DescriptorProtos.FileOptions.JAVA_MULTIPLE_FILES_FIEL
...
 import kotlin.reflect.KClass
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 
-class `'CompilerEvents' should` {
+
+@DisplayName("`CompilerEvents` should")
+class CompilerEventsSpec {
```
armiol, line 74: Maybe, both `CompilerEvents` and this test should be in `.event` package?

armiol (approved):
@alexander-yevsyukov LGTM with some comments to consider.

compiler/src/test/kotlin/io/spine/protodata/CompilerEventsSpec.kt
alexander-yevsyukov: Good catch. There were a couple of other tests too.

build.gradle.kts
alexander-yevsyukov: Good idea. Done.

build.gradle.kts
```diff
@@ -104,100 +104,41 @@ allprojects {
             force(
                 io.spine.internal.dependency.Grpc.protobufPlugin,
```
yevhenii-nadtochii, line 105: We should be able to import `Grpc` object here instead of FQN.

compiler/src/main/kotlin/io/spine/protodata/AstExtension.kt
```diff
@@ -26,13 +26,15 @@
 
 @file:JvmName("Ast")
 
+@file:Suppress("TooManyFunctions")
```
yevhenii-nadtochii, line 29: This rule shouldn't be reported for `*Extensions.kt` files (if this is Detekt's one). I suggest fix the name of the file to have plural ending.

compiler/src/main/kotlin/io/spine/protodata/AstExtension.kt
```diff
@@ -112,7 +114,7 @@ public fun Service.typeUrl(): String = name.typeUrl()
  * Shows if this field is a `map`.
  *
  * If the field is a `map`, the `Field.type` contains the type of the value, and
- * the `Field.map.key_type` contains the type the the map key.
+ * the `Field.map.key_type` contains the type the map key.
```
yevhenii-nadtochii, line 117: "contains the type **of** the map key"

compiler/src/main/kotlin/io/spine/protodata/AstExtension.kt
alexander-yevsyukov: @yevhenii-nadtochii, thanks for these. I'll address them in the following PR.

PR #108, opened 2022-11-10
Apply Detekt and address its findings

yevhenii-nadtochii:
This PR moves Detekt code analysis from a workflow to Gradle build.

The reported findings were addressed.

Significant ones:

1. `Run.allowMissingDirs` option has been dropped. It was actually unused. ProtoData always ignores missing source directories when multiple ones are passed. The description of `sourceDirs` option has been updated to include this information in its `help` message.
2. `AstExtension.kt` was renamed to `AstExtensions.kt` because `*Extensions.kt` files are not checked for `TooManyFunctions` rule.

The library version is bumped to `0.4.4`.


yevhenii-nadtochii:
@alexander-yevsyukov @armiol Please take a look.

armiol:
@yevhenii-nadtochii this is too much:
```
SourceFileKt.at(sources.file(javaFilePath), classScope)
            .withExtraIndentation(INDENT_LEVEL)
            .add(lines);
```

It was much more readable previously, without this "Kt" ending. 

Can we deal with it somehow, or revert?

cli/src/main/kotlin/io/spine/protodata/cli/Main.kt
```diff
@@ -84,6 +84,7 @@ private fun readVersion(): String = Version.fromManifestOf(Run::class.java).valu
  * Finally, the renderers apply required changes to the source set with the root path, supplied in
  * the `--source-root` parameter.
  */
+@Suppress("TooManyFunctions") // It is OK for the main command.
```
armiol, line 87: I would say, "for `main` entry point."

tests/protodata-extension/src/main/java/io/spine/protodata/test/annotation/AnnotationRenderer.java
```diff
@@ -72,9 +73,9 @@ private void renderFor(Annotated field, SourceFileSet sourceSet) {
             return;
         }
 
-        sourceSet.file(path)
-                 .at(getter)
-                 .withExtraIndentation(INDENT_LEVEL)
-                 .add('@' + field.getJavaAnnotation());
+
+        SourceFileKt.at(sourceSet.file(path), getter)
```
armiol, line 77: See my PR-level comment on this one.

armiol (changes requested):
Please see my comments.

tests/protodata-extension/src/main/java/io/spine/protodata/test/annotation/AnnotationRenderer.java
armiol: Let's go for the second option.

armiol: It's not about Java compatibility per se. It's about the language and readability when used in Java.

alexander-yevsyukov:
> Run.allowMissingDirs 

@yevhenii-nadtochii, please check it with @dmdashenkov. I've seen this flag in command line parameters, but cannot say if it's really used.

alexander-yevsyukov (approved):
LGTM

dmdashenkov:
@yevhenii-nadtochii, seems like ProtoData now operates as if that flag is always set to `true`. I think we can safely delete it. Please take a look at the descriptions of CLI properties that specify directory paths and make sure that the help messages are in line with the actual behaviour. 

Please also check if the ProtoData Gradle plugin uses this option. It does in my branch (`LaunchProtoData.kt:173`). If so, please remove it there.

compiler/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -281,6 +247,38 @@ internal constructor(
...
+ * Creates a new fluent builder for adding code at the given [insertionPoint].
+ *
+ * If the [insertionPoint] is not found in the code, no action will be performed as the result.
+ * If there are more than one instances of the same insertion point, the code will be added to
+ * all of them.
+ *
+ * Insertion points should be marked with comments of special format. The added code is always
+ * inserted after the line with the comment, and the line with the comment is preserved.
+ */
+public fun SourceFile.at(insertionPoint: InsertionPoint): SourceAtPoint =
```
dmdashenkov, line 267: Regarding `at(..)` and `lines()`, I think you should reconsider. Those are not auxiliary functions, but the primary API any SPI-user will be using. If there are some formal requirements regarding what is considered "auxiliary" and what is not, let's see if it should be changed or if we should suppress these particular cases.

yevhenii-nadtochii:
@armiol @dmdashenkov Please take a look.

I've restored `SourceFile` and suppressed the warning. 

`LaunchProtoData` was updated not to pass the deleted option, and the help message for `sourceSirs` option now mentions that non-existent directories are ignored.

cli/src/main/kotlin/io/spine/protodata/cli/Main.kt
```diff
@@ -356,16 +340,38 @@ internal class Run(version: String) : CliktCommand(
         try {
             return createByName(className, classLoader)
         } catch (e: ClassNotFoundException) {
-            error(e.message)
-            error("Please add the required class `$className` to the user classpath.")
+            printError(e.message)
```
armiol, line 343: In this throwable a stacktrace is also very valuable. Can we please print it as well?

armiol (approved):
LGTM with a comment to address.

PR #112, opened 2022-11-18
Improve modularity

alexander-yevsyukov:
This PR improves modularity introducing `cli-api` and `api` modules thus separating API modules from implementation.

Further analysis of API visibility and possible improvements to package structure would be made in the following PRs.

The triggers for introducing this change were:
   1) `help` fields for the command line parameters that mention classes that belonged to another module. Because of that the class names could not be referenced directly and had to be strings.
   2) command line parameter names that were referenced, again, as strings from the `plugin` module.
These were indicators that we have implicit dependencies that were masked by string parameters.

In addition, we had a dependency issues in Gradle that we resolved via introducing fat JARs. This PR is another step towards improving the way we develop and debug ProtoData.


alexander-yevsyukov:
> expand the PR description with more detail on why we need this increasing module granularity.

@armiol, done. Do you want me to add more or other details?

armiol (approved):

PR #119, opened 2023-01-12
Migrate to Protobuf Gradle Plugin v0.9.1

alexander-yevsyukov:
This PR migrates ProtoData to be built and use Protobuf Gradle Plugin v0.9.1.

Other notable changes:
  * The names of the `LaunchProtoData` are now calculated using standard Gradle conventions via infix form like `launchProtoData` for the `main` source set and `launchSourceSetNameProtoData` for the rest.
  
     The `BuildExtensions.kt` files under `buildSrc` will be adjusted correspondingly in `config` after sub-projects depending on ProtoData will be migrated to the version introduced by this PR (`0.6.0`).
     
  * The `Project.sourceSets` extension property introduced in the `io.spine.protodata.gradle.plugin` package was removed in favor of the similar property from the `io.spine.tools.gradle.project` package of `tool-base`.

### DIFF
--- a/gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/LaunchProtoData.kt
+++ b/gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/LaunchProtoData.kt
@@ -35,6 +35,7 @@ import io.spine.protodata.cli.RequestParam
 import io.spine.protodata.cli.SourceRootParam
 import io.spine.protodata.cli.TargetRootParam
 import io.spine.protodata.cli.UserClasspathParam
+import io.spine.tools.gradle.protobuf.ProtobufDependencies.sourceSetExtensionName
 import java.io.File.pathSeparator
 import org.gradle.api.Action
 import org.gradle.api.Task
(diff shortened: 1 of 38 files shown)


gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/LaunchProtoData.kt
```diff
@@ -190,3 +193,31 @@ private fun Provider<List<Directory>>.absoluteDirs() = takeIf { it.isPresent }
...
+ * Tells if the request file for this task exists.
+ *
+ * Logs error if the given source set contains `proto` directory which contains files,
+ * which assumes that the request file should have been created.
+ */
+internal fun LaunchProtoData.checkRequestFile(sourceSet: SourceSet): Boolean {
+    val requestFile = requestFile.get().asFile
+    if (!requestFile.exists() && sourceSet.containsProtoFiles()) {
+        project.logger.error(
+            "The task `${name}` was skipped because" +
```
armiol, line 207: "Task ... was skipped" sounds not as an error description, but rather like a warning. 

I suggest starting with something like "Error locating the request file...".

armiol (approved):
LGTM with a suggestion.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/LaunchProtoData.kt
armiol: @alexander-yevsyukov I think you might have missed this suggestion.

alexander-yevsyukov: Yes, you're right. Was reviewing from the mobile app, and it wasn't visible. I'm going to address it in a next PR.

PR #121, opened 2023-02-05
Improve codegen for annotations

alexander-yevsyukov:
This PR improves API for annotating Java types. Namely:

  * An abstract class `TypeAnnotation` was introduced. Its constructor accepts a class of `Annotation` to be used in the generated code.
  
  * `GeneratedGenerator` class was renamed to `GeneratedTypeAnnotation` and placed under the `...java.annotation` package. The class is `open` so that users can create parameterless descendants, as it's required by ProtoData conventions.
  
  * `SuppressRenderer` class was renamed to `SuppressWarningsAnnotation` and placed under `...java.annotation` package.
  
  * `SuppressConfig` proto type was renamed to `SuppressionSettings`. The previous name sounded as an imperative, which was confusing.

#### Other changes:
  * Kotlin proto DSL was applied in `AstExtensions.kt`.
  * `target: Language` property was changed from `private` to `protected` in `InsertionPointPrinter`. Derived classes may need to know for which language they work.
  * The `InsertionPointPrinter.render()` was refactored extracting two extension `fun` for `SourceFile` and `InsertionPoint`.


armiol (approved):
LGTM with some comments that we have already addressed during our vocal conversation.

PR #126, opened 2023-03-03
No target for built-in annotations

dmdashenkov:
In JDK 19, the `SuppressWarnings` annotation has no `@Target` and is resolved by the compiler/JRE implicitly instead. This causes our code generator to fail when adding that annotation to the generated code.

See the [JDK issue](https://bugs.openjdk.org/browse/JDK-8280745).


armiol (approved):
@dmdashenkov LGTM. I would also create an issue on our side, just to remember to revisit this problem in the future (maybe, they will fix the bug?!).

alexander-yevsyukov (approved):
LGTM

dmdashenkov:
@armiol, this is in JDK 19 release, so this will never be fully reversed. It's also in the specification, so, technically, not a bug.

PR #127, opened 2023-03-17
Emit events for all known types

dmdashenkov:
In this PR we enable ProtoData plugin authors to listen to Protobuf compiler events that describe types from dependencies.

Previously, events like `TypeEntered` were only emitted for Proto definitions from within the module that ProtoData is processing. This is good for many cases, since this removes the need to check whether or not any code has to be generated based on a given event. However, this limited the users to only the types defined in the module. There was no way to know about a Proto message type from outside the module.

Now, we introduce a new `DepdendencyDiscovered` event and a `ProtobufDependencyFile` view. The two new types allow ProtoData plugin authors to generate code with awareness of the module's dependencies.


dmdashenkov:

This is a small yet consequential change set. Let's further discuss it a video call if necessary.

.github/workflows/build-on-ubuntu.yml
```diff
@@ -22,23 +22,13 @@ jobs:
...
-      - name: Publish Unit Test Report
+      - name: Publish Test Report
         uses: mikepenz/action-junit-report@v3.5.2
         if: always() # always run even if the previous step fails
         with:
-          check_name: Regular Unit Tests
           report_paths: '**/build/test-results/test/TEST-*.xml'
           require_tests: true # will fail workflow if test reports not found
 
-      # See: https://github.com/marketplace/actions/junit-report-action
```
alexander-yevsyukov, line 33: Can we revert these changes please? Separation of test reports was needed some time ago when investigating build failures at CI.

api/src/main/kotlin/io/spine/protodata/event/EnumCompilerEvents.kt
```diff
@@ -68,13 +68,15 @@ internal class EnumCompilerEvents(
             EnumEntered.newBuilder()
```
alexander-yevsyukov, line 68: Please consider using Kotlin DSL for Protobuf. It should be already available here. I've just tried it in another branch and it worked for me.

api/src/main/kotlin/io/spine/protodata/event/CompilerEvent.kt
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Shows if this event should result in code generation.
+     *
+     * If `false`, the event is purely informative and should not add any code generation tasks.
+     * Such events are emitted for the Proto definitions that come from a module's dependency.
+     *
+     * If `true`, the event describes a Proto definition that belongs to the currently analyzed
+     * module and thus may result in code generation.
+     */
+    public fun getGenerationRequested(): Boolean
```
alexander-yevsyukov, line 45: If we take all the `CompilerEvent`s, what percentage of them would be non-generative?

I'm asking to see if we may way to make it vice-versa, e.g. have `isInformative()`/`getInformative()` instead.

alexander-yevsyukov: This way could probably reduce the number of setting the flag for enabling code generation.

alexander-yevsyukov (changes requested):
Please see my comments. In general, I think that the `generation_requested` flag approach should be discussed.

api/src/main/kotlin/io/spine/protodata/event/CompilerEvent.kt
dmdashenkov: When setting the flag, it's not relevant if the value is `true` or `false`, the number of operations will be the same. As per the percentage, I'd assume that on bigger projects `true` would win out over `false`. But it's impossible to tell in the general case.

armiol (changes requested):
@dmdashenkov I am now requesting changes, so that you could re-request from me, once this PR is ready. Alexander is off for a while, so it will be just me.

dmdashenkov:
@armiol, PTAL.

api/src/main/proto/spine/protodata/events.proto
```diff
@@ -264,3 +265,15 @@ message RpcExited {
 
     RpcName rpc = 3;
 }
+
+// Emitted when a Protobuf dependency file is discovered.
```
armiol, line 269: Is this about discovering an `import` in our code? I am not sure what do you mean by "discovered".

api/src/main/proto/spine/protodata/events.proto
```diff
@@ -264,3 +265,15 @@ message RpcExited {
...
+//
+// Normally, no code should be generated for dependencies. However, they can be used for additional
+// info when generating code for the module's own types.
+//
+message DependencyDiscovered {
```
armiol, line 274: I would be more precise with this name, so that Java- and Proto-level dependencies could be distinguished.

compiler/src/main/kotlin/io/spine/protodata/backend/AstComponents.kt
```diff
@@ -0,0 +1,198 @@
...
+        .setDeclaringType(declaringType)
+        .setNumber(number)
+        .setOrderOfDeclaration(index)
+        .assignTypeAndCardinality(this)
+        .setDoc(documentation.forField(this))
+        .build()
+}
+
+/**
+ * Assigns the field type and cardinality (`map`/`list`/`oneof_name`/`single`) to the receiver
```
armiol, line 79: Is this, in fact, copying the field type and the cardinality? If so, maybe we could get rid of this "assign" term which feels a bit over-used in our domain?

compiler/src/main/kotlin/io/spine/protodata/backend/AstComponents.kt
```diff
@@ -0,0 +1,198 @@
...
+ * builder.
+ *
+ * @return the receiver for method chaining.
```
armiol, line 82: I think, in Kotlin we still don't put periods in `@return` statement descriptions.

compiler/src/main/kotlin/io/spine/protodata/backend/AstComponents.kt
```diff
@@ -0,0 +1,198 @@
...
+ * Extracts metadata from this file descriptor, excluding file options.
+ *
+ * @see toFileWithOptions
+ */
+internal fun Descriptors.FileDescriptor.toFile() = file {
+    path = path()
+    packageName = `package`
+    syntax = this@toFile.syntax.toSyntaxVersion()
+}
+
```
armiol, line 198: I am not sure whether we can remove this empty line (and not break EOF), but if we can, let's.

compiler/src/main/kotlin/io/spine/protodata/backend/CompilerEvents.kt
```diff
@@ -46,12 +46,14 @@ public object CompilerEvents {
      *
      * The resulting sequence is always finite, it's limited by the type set.
      */
-    public fun parse(request: CodeGeneratorRequest): Sequence<EventMessage> {
+    public fun parse(request: PluginProtos.CodeGeneratorRequest): Sequence<EventMessage> {
```
armiol, line 49: I am not sure why we now include the reference to the parent type. Please review.

compiler/src/main/kotlin/io/spine/protodata/backend/CompilerEvents.kt
```diff
@@ -46,12 +46,14 @@ public object CompilerEvents {
...
         val filesToGenerate = request.fileToGenerateList.toSet()
         val files = FileSet.of(request.protoFileList)
         return sequence {
-            files.files()
-                .filter { it.name in filesToGenerate }
+            val (ownFiles, dependencies) = files.files()
```
armiol, line 53: Maybe, that's something which should be reflected in the description for this method?

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
```diff
@@ -0,0 +1,191 @@
...
+import io.spine.protodata.path
+import io.spine.protodata.service
+import io.spine.protodata.typeUrl
+
+/**
+ * Creates a `DependencyDiscovered` event from the given file descriptor.
+ *
+ * The event reflects all the definitions from the file.
+ */
+internal fun toDependencyEvent(fileDescriptor: Descriptors.FileDescriptor) =
```
armiol, line 49: How about `discoverDependencies(fileDescriptor: Descriptors.FileDescriptor)`?

compiler/src/main/kotlin/io/spine/protodata/backend/Documentation.kt
```diff
@@ -24,32 +24,37 @@
...
+import com.google.protobuf.Descriptors.FileDescriptor
 import com.google.protobuf.Descriptors.MethodDescriptor
 import com.google.protobuf.Descriptors.OneofDescriptor
 import com.google.protobuf.Descriptors.ServiceDescriptor
+import io.spine.protodata.Doc
 import io.spine.string.trimWhitespace
 import io.spine.util.interlaced
 
 /**
  * Documentation contained in a Protobuf file.
```
armiol, line 44: Given its description, I don't really understand why such a generic thing is now in `.backend`.

compiler/src/main/kotlin/io/spine/protodata/backend/Documentation.kt
```diff
@@ -149,10 +154,10 @@ private constructor(private val value: List<Int>) {
          */
         fun fromMessage(descriptor: Descriptor): LocationPath {
             val numbers = mutableListOf<Int>()
-            numbers.add(FileDescriptorProto.MESSAGE_TYPE_FIELD_NUMBER)
+            numbers.add(DescriptorProtos.FileDescriptorProto.MESSAGE_TYPE_FIELD_NUMBER)
```
armiol, line 157: There is a lot of such changes in this PR. Don't you think it's less convenient to address the types and their members by specifying a "full" (so to speak) name, rather than stripping some common parts by having them "statically" imported?

armiol (changes requested):

alexander-yevsyukov (changes requested):
Please see my comments.

api/src/main/proto/spine/protodata/events.proto
dmdashenkov: I've updated the doc somewhat. I'd like to avoid adding the `Proto` prefix because: 
 - we only work with Protobuf, no Java or JS or other dependency is ever considered in ProtoData (not talking about the Gradle compiler here, only ProtoData proper);
 - this is one of a series of events such as `TypeEntered`, `OptionDiscovered`, etc., which do not specify the language eiher.

compiler/src/main/kotlin/io/spine/protodata/backend/CompilerEvents.kt
dmdashenkov: I'm assuming that you thought that `files()` returned a tuple. In actuality, it's the `partition(..)` that chains to this call. I've updated the code layout so that it's more clear.

compiler/src/main/kotlin/io/spine/protodata/backend/AstComponents.kt
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * The resulting [Field] will not reflect the field options.
+ *
+ * @see buildFieldWithOptions
+ */
+internal fun Descriptors.FieldDescriptor.buildField(
+    declaringType: TypeName,
+    documentation: Documentation
+): Field {
+    return Field.newBuilder()
```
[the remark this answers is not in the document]
dmdashenkov: In methods that already have a receiver, I found it easier to use the builder rather than the DSL. Otherwise, we'd have to specify which `this` we're talking about each time:
```kotlin
this@field.number = this@buildField.number
```

It's similar in other places in `AstComponents`.

compiler/src/main/kotlin/io/spine/protodata/backend/ServiceCompilerEvents.kt
```diff
@@ -107,15 +102,15 @@ internal class ServiceCompilerEvents(
             RpcOptionDiscovered.newBuilder()
```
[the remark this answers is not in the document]
dmdashenkov: I'd like to postpone such a transition to a separate changeset. If we've decided that Kotlin DSL is the standard, let's have a separate PR that will migrate the whole codebase.

compiler/src/test/kotlin/io/spine/protodata/backend/CodeGenerationContextSpec.kt
```diff
@@ -27,72 +27,156 @@
...
+            assertSourceFile
+                .exists()
+            val actual = assertSourceFile.actual()!!.state() as ProtobufSourceFile
+
+            val types = actual.typeMap
+            val typeName = "type.spine.io/spine.protodata.test.Journey"
+            assertThat(types)
+                .containsKey(typeName)
+            val journeyType = types[typeName]!!
+            assertThat(journeyType.name.typeUrl())
```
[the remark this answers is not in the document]
dmdashenkov: I have seen the new method of writing assertions, but am not sure. Should we mix `ProtoTruth` with KoTest in the same test case?

dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL at the changes and my comments.

api/src/main/proto/spine/protodata/source.proto
```diff
@@ -70,3 +70,16 @@ message ProtobufSourceFile {
     //
     map<string, Service> service = 5;
 }
+
+// A Protobuf definitions file which is included into the module as a dependency.
+//
+// Unlike `ProtobufSourceFile`, `ProtobufDependencyFile` should not result in any code generation.
+// Rather, it serves an informative purpose.
+//
+message ProtobufDependencyFile {
```
alexander-yevsyukov, line 79: I think the suffix `File` in this type and in `ProtobufSourceFile` makes it a bit harder to understand the difference. More, we have the `file` property under `ProtobufSourceFile` and `content` under `ProtobufDependencyFile`, which makes it a bit more confusing. In fact, the dependency is on the file. We just don't generate the code after it.

api/src/main/proto/spine/protodata/source.proto
```diff
@@ -42,7 +42,7 @@ import "spine/protodata/ast.proto";
 // Includes all the message types declared in the file, along with their fields.
 //
 message ProtobufSourceFile {
-    option (entity).kind = PROJECTION;
+    option (entity).kind = VIEW;
 
     FilePath file_path = 1;
```
alexander-yevsyukov, line 47: We probably need to document the id fields here saying that its value should be equal to one from the `file.path` property.

compiler/src/main/kotlin/io/spine/protodata/backend/AstComponents.kt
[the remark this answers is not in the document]
alexander-yevsyukov: We do have name clashes in some cases, but they are rare. I've changed several such blocks myself and know what you're talking about.

Try to apply the DSL to this particular case and you'll see that the code becomes much more compact and easier to read.

api/src/main/proto/spine/protodata/source.proto
```diff
@@ -70,3 +70,16 @@ message ProtobufSourceFile {
...
+//
+// Unlike `ProtobufSourceFile`, `ProtobufDependencyFile` should not result in any code generation.
+// Rather, it serves an informative purpose.
+//
+message ProtobufDependencyFile {
+    option (entity).kind = VIEW;
+
+    FilePath file_path = 1;
+
+    ProtobufSourceFile content = 2;
```
alexander-yevsyukov, line 84: Can we have it named as `file` instead, please?

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
```diff
@@ -0,0 +1,194 @@
...
+                .toMap()
+        )
+        result.putAllEnumType(
+            enumTypes()
+                .map { it.name.typeUrl() to it }
+                .toMap()
+        )
+        result.putAllService(
+            services()
+                .map { it.name.typeUrl() to it }
```
alexander-yevsyukov, line 78: Please extract the `.map {...}.toMap()`. We have them 3 times already here. Please also look for similar cases and refactor.

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Builds the service definitions.
+     *
+     * @return all the services declared in the file, including the nested ones.
+     */
+    fun FileDescriptor.services(): Sequence<Service> =
+        services.asSequence().map { it.asService() }
+
+    private fun Descriptor.asMessage() = messageType {
```
alexander-yevsyukov, line 130: This method deserves documentation. It looks complicated. Please also consider refactoring.

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
```diff
@@ -0,0 +1,194 @@
...
+    return result.build()
+}
+
+/**
+ * A builder for the Protobuf definitions of a single `.proto` file.
+ *
+ * @property path the relative file path to the Protobuf file
+ * @property documentation all the documentation and comments present in the file
+ */
+private class DefinitionsBuilder(
```
alexander-yevsyukov, line 91: This class forms context with `path` and `documentation`, and extension functions it exposes. It's then used once in the `with` clause making it easier to traverse in the types.

But the purpose of this class should be documented, at least briefly. Otherwise, this somewhat complex piece of code requires “reverse” engineering.

Please also consider using a different name, avoiding the `Builder` suffix.

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
```diff
@@ -0,0 +1,194 @@
...
+    private fun Descriptor.asMessage() = messageType {
+        val typeName = name()
+        name = typeName
+        file = path
+        doc = documentation.forMessage(this@asMessage)
+        option.addAll(listOptions(options))
+        if (containingType != null) {
+            declaredIn = containingType.name()
+        }
+        oneofGroup.addAll(realOneofs.map { oneofGroup {
```
alexander-yevsyukov, line 139: This block is hard to understand. Please consider refactoring it.

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
```diff
@@ -0,0 +1,194 @@
...
+            val groupName = it.name()
+            name = groupName
+            field.addAll(it.fields.map { f -> f.buildFieldWithOptions(typeName, documentation) })
+            option.addAll(listOptions(options))
+            doc = documentation.forOneof(it)
+        }
+        })
+        field.addAll(fields.map { it.buildFieldWithOptions(typeName, documentation) })
```
alexander-yevsyukov, line 147: We have a this piece of code repeated: `fields.map { it.buildFieldWithOptions(typeName, documentation) }`.

Can we refactor so that we tell it in a bit more concise manner?

compiler/src/main/kotlin/io/spine/protodata/backend/MessageCompilerEvents.kt
```diff
@@ -24,17 +24,23 @@
...
 import com.google.protobuf.Descriptors
-import com.google.protobuf.Empty
 import io.spine.base.EventMessage
-import io.spine.protodata.Documentation
-import io.spine.protodata.Field
 import io.spine.protodata.File
-import io.spine.protodata.MessageType
 import io.spine.protodata.OneofGroup
 import io.spine.protodata.TypeName
+import io.spine.protodata.event.FieldEntered
```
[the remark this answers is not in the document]
alexander-yevsyukov: @dmdashenkov, we do need `event` package so that Javadocs, KDocs, and source code JARs are easier to understand. Please do have events coming under the `event` package.

compiler/src/test/kotlin/io/spine/protodata/backend/CodeGenerationContextSpec.kt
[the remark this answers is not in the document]
alexander-yevsyukov: Yes, please do. Kotest assertions are much easier to read. We'll migrate Protobuf-related assertions later.

alexander-yevsyukov (changes requested):
Please see my comments and answers.

compiler/src/main/kotlin/io/spine/protodata/backend/DependencyEvents.kt
alexander-yevsyukov: How about `discoverDependencies(fd: FileDescriptor)`? :)

api/src/main/kotlin/io/spine/protodata/ProtoDeclaration.kt
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public val name: ProtoDeclarationName
+
+    /**
+     * Obtains the type URl of the type.
+     *
+     * A type URL contains the type URL prefix and the qualified name of the type separated by
+     * the slash (`/`) symbol. See the docs of `google.protobuf.Any.type_url` for more info.
+     */
+    public fun typeUrl(): String? {
```
alexander-yevsyukov, line 50: If the return value can really be `null`, please document when such a case can occur.

compiler/src/main/kotlin/io/spine/protodata/backend/AstComponents.kt
```diff
@@ -0,0 +1,197 @@
...
+internal fun FieldDescriptor.buildField(
+    declaringType: TypeName,
+    documentation: Documentation
+) = field {
+    name = name()
+    orderOfDeclaration = index
+    doc = documentation.forField(this@buildField)
+    this@field.number = this@buildField.number
+    this@field.declaringType = declaringType
+    copyTypeAndCardinality(this@buildField)
```
alexander-yevsyukov, line 85: Please extract `this@buildField` into a variable (e.g. `fieldDescr`, or `thisField`). This way you'd have fewer “dogs” around.

compiler/src/main/kotlin/io/spine/protodata/backend/ServiceCompilerEvents.kt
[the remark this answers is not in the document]
alexander-yevsyukov: OK, no problem. Let's migrate it gradually.

alexander-yevsyukov (changes requested):
Please see my additional comments.

api/src/main/proto/spine/protodata/source.proto
alexander-yevsyukov: Added [an issue](https://github.com/SpineEventEngine/validation/issues/83) for the `same_as` option.

api/src/main/kotlin/io/spine/protodata/ProtoDeclaration.kt
dmdashenkov: Oops, that was a mistake. Fixed it.

dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL again.

api/src/main/kotlin/io/spine/protodata/ProtoDeclaration.kt
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public val name: ProtoDeclarationName
+
+    /**
+     * Obtains the type URL of the type.
+     *
+     * A type URL contains the type URL prefix and the qualified name of the type separated by
+     * the slash (`/`) symbol. See the docs of `google.protobuf.Any.type_url` for more info.
+     */
+    public fun typeUrl(): String {
```
alexander-yevsyukov, line 50: Why don't we have it as another property rather than a function? Just curious, how you see it.

One reason I can think of is to have shorter names in Java code. But we aim to Kotlin anyway, and we always can have `@JvmName` defined, should we want to.

compiler/src/test/kotlin/io/spine/protodata/backend/event/CompilerEventsSpec.kt
```diff
@@ -183,8 +201,8 @@ class CompilerEventsSpec {
 
         event.option.name shouldBe  "idempotency_level"
         event.option.value.unpackGuessingType() shouldBe enumValue {
-            name = IdempotencyLevel.NO_SIDE_EFFECTS.name
-            number = IdempotencyLevel.NO_SIDE_EFFECTS_VALUE
+            name = DescriptorProtos.MethodOptions.IdempotencyLevel.NO_SIDE_EFFECTS.name
```
alexander-yevsyukov, line 204: Can we import `MethodOptions`, please?

alexander-yevsyukov (changes requested):
Please see my comments. Almost there.

dmdashenkov:

PR #129, opened 2023-04-04
Restore binary backward compatibility

dmdashenkov:
In this PR we bring back a few methods removed in a previous change set.

The methods are all extension methods that were replaced by more fitting constructs. However, we still need them for backward compatibility.

Note. Only binary compatibility is guaranteed. The methods will not show up at compile-time, since the migration is fairly simple.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
+++ b/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
@@ -45,6 +45,38 @@ import io.spine.protodata.CallCardinality.SERVER_STREAMING
 import io.spine.protodata.CallCardinality.UNARY
 
 /**
+ * Obtains the type URl of the type.
+ *
+ * A type URL contains the type URL prefix and the qualified name of the type separated by
+ * the slash (`/`) symbol. See the docs of `google.protobuf.Any.type_url` for more info.
+ *
+ * @see MessageType.qualifiedName
+ * @see TypeName.typeUrl
+ */
+@Deprecated(
+    message = "Replaced with a Kotlin property (or a Java-style getter).",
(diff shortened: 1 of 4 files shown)


dmdashenkov:
@armiol, PTAL.

api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
```diff
@@ -44,6 +44,38 @@ import io.spine.protodata.CallCardinality.CLIENT_STREAMING
...
+ */
+@Deprecated(
+    message = "Replaced with a Kotlin property (or a Java-style getter).",
+    replaceWith = ReplaceWith("typeUrl"),
+    level = DeprecationLevel.HIDDEN
+)
+public fun MessageType.typeUrl(): String = typeUrl
+
+/**
+ * Obtains the type URl of the type.
```
armiol, line 64: `URL`.

api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
```diff
@@ -61,6 +93,47 @@ public fun TypeNameOrBuilder.qualifiedName(): String {
...
+ */
+@Deprecated(
+    message = "Replaced with a Kotlin property (or a Java-style getter).",
+    replaceWith = ReplaceWith("typeUrl"),
+    level = DeprecationLevel.HIDDEN
+)
+public fun ServiceName.typeUrl(): String = typeUrl
+
+/**
+ * Obtains the type URl of this service.
```
armiol, line 126: Same as above.

armiol (approved):
@dmdashenkov LGTM with a couple of comments to address.

PR #130, opened 2023-04-10
Log if configuration failed

dmdashenkov:
If an attempt to configure the Kotlin compilation task fails, we log the name and the type of the task for easier debugging.


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #131, opened 2023-04-10
Support Protobuf Gradle plugin v0.9.2

dmdashenkov:
In this PR we make a change necessary for the ProtoData Gradle plugin to work with the Protobuf Gradle plugin of version 0.9.2.

Previously, we overrode `generatedBaseDir`, which resulted in the Proto-generated files being added to compile classpath twice. Now, we rely on the default value, thus skipping extra file copying operations and ensuring only one version of the generated files exists.

The ProtoData Gradle plugin will still modify the `compileKotlin` and `compileJava` tasks' classpath. With this change, the mechanism will work just fine again.

ProtoData itself does not yet use the new version of the plugin yet. We'll be able to migrate it after the `mc-java` is migrated.

### DIFF
--- a/gradle-plugin/src/functionalTest/kotlin/io/spine/protodata/gradle/plugin/PluginSpec.kt
+++ b/gradle-plugin/src/functionalTest/kotlin/io/spine/protodata/gradle/plugin/PluginSpec.kt
@@ -26,7 +26,6 @@
 
 package io.spine.protodata.gradle.plugin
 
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

Note: this is an experiment, rather than a guaranteed solution, but looks promising to me.

alexander-yevsyukov (approved):
LGTM

PR #134, opened 2023-05-22
Fix Gradle config for `new-gradle-proto-plugin`

dmdashenkov:
In this PR we fix the Gradle configuration issues in `new-gradle-proto-plugin` that lead to duplicate classes.

Long story short, we also needed to bump `mc-java` along with `ProtoData`.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -57,12 +57,12 @@ dependencies {
 /**
  * Force `generated` directory and Kotlin code generation.
  */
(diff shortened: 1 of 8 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL with your local setup. For me, there are still some failing tests, but the duplicate files issue is resolved with no extra configuration.

alexander-yevsyukov (approved):
LGTM

PR #140, opened 2023-05-29
Add logging level command line parameters

alexander-yevsyukov:
This PR extends ProtoData CLI with `--debug` and `--info` command line parameters that enable corresponding logging levels.

### Note on testing console output
Testing of console output using [System Lambda](https://github.com/stefanbirkner/system-lambda) turned to be fragile. It worked as expected when running from IDEA, but failed when running under Gradle. Assuming that System Lambda is not compatible with Java 16, no further attempts to find out the cause of instability were made. Tapping console output in a POJ way was attempted, which also proved to be unreliable when working under Gradle. In the end, a pseudo-`Renderer` was created which tests the logging level from inside ProtoData.

Even though the issue with console output checks was solved for this PR, we still need to find out a good way for verifying the console output for this and other modules of the Spine SDK.


README.md
```diff
@@ -62,3 +62,9 @@ Start using the app:
 
 Note that it is necessary to keep the structure of the archive intact: the scripts in `bin` rely
 on the relative path to other files in the archive.
+
+[ubuntu-build]: https://github.com/SpineEventEngine/ProtoData/actions/workflows/build-on-ubuntu.yml
```
alexander-yevsyukov, line 66: Jekyll Front Matter is not supported for README.md files under GitHub. So we cannot have a page-wide variables followed by Liquid expansion.

cli-api/src/main/kotlin/io/spine/protodata/cli/CommandLineInterface.kt
```diff
@@ -161,6 +161,22 @@ public object ConfigFormatParam : Parameter(
...
 
+public object InfoLoggingParam : Parameter(
+    name = "--info",
+    shortName = "-I",
+    help = """
+        Set log level to `INFO`.        
+    """
+)
+
+public object DebugLoggingParam : Parameter(
```
dmdashenkov, line 172: Please add short descriptions for these new objects.

dmdashenkov (approved):
LGTM with one comment.

PR #141, opened 2023-05-30
Simplify working with `OptionsProvider`s, removal of deprecated API

alexander-yevsyukov:
This PR simplifies the way ProtoData works with custom `OptionsProvider`s making it rely on the convention introduced by the `OptionsProvider` interface. It assumes that all the implementations can be loaded by `ServiceLoader`.

Also this PR removes the deprecated API. Hence the version bumped to `0.9.0`.

PR #144, opened 2023-06-13
Restore option autoloading

dmdashenkov:
In a [previous PR](#141), we have removed the ability to specify option providers by name. With this, however, we have also, unintentionally, removed the ability of ProtoData to load custom options from the sources being processed.

In this PR we restore such ability. Now, just like before v0.9.0, ProtoData automatically loads options from all the Proto files in the source file set.


dmdashenkov:

cli/src/main/kotlin/io/spine/protodata/cli/app/FileSetOptionsProvider.kt
```diff
@@ -24,21 +24,32 @@
...
-import io.spine.protodata.test.MetaProto;
+import com.google.protobuf.ExtensionRegistry
+import io.spine.code.proto.FileSet
+import io.spine.option.OptionsProvider
+import io.spine.protobuf.outerClass
+import io.spine.protobuf.registerAllExtensions
 
 /**
- * Registers custom proto options introduced by this library.
+ * An [OptionsProvider] which provides all the options defined in a set of Protobuf files.
```
armiol, line 36: As discussed vocally, please extend the description of this provider, so that next time we could understand the use-case faster.

armiol (approved):
@dmdashenkov LGTM with a comment to address.

PR #145, opened 2023-07-01
Bump internal dependencies

alexander-yevsyukov:
This PR bumps dependency on `logging`, `core-java`, `validation`, and other internal dependencies. Also latest `config` was applied.

PR #146, opened 2023-07-01
Fix and trigger publishing

alexander-yevsyukov:
This PR reverts the changes to `PublishingExts.kt` [applied recently](https://github.com/SpineEventEngine/ProtoData/commit/56dd5259718f029a21f961176249fec6d6cd5812#diff-5030e40d04de8d6aa565900eec666939e6811ab5521087d90668024a6da2f262) in the hope to fix the publishing which did not start when merging the [previous PR](https://github.com/SpineEventEngine/ProtoData/pull/145).

### DIFF
--- a/.github/workflows/build-on-ubuntu.yml
+++ b/.github/workflows/build-on-ubuntu.yml
@@ -4,6 +4,7 @@ on: push
 
 jobs:
   build:
+    name: Build under Ubuntu
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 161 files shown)


armiol (approved):

alexander-yevsyukov:
@armiol, I've removed `Build under Windows` from the status checks because the checks starts on `pull_request`, while `Build under Ubuntu` is started on `push`.

My guess is that when pushing to `master` the `Build under Windows` is never started. Previously, the whole arrangement worked because we didn't have a title for a workflow. It was always `build` by the name of a job (whatever came first worked).

PR #151, opened 2023-07-20
Move `TypeSystem` from `Validation`

dmdashenkov:
In this PR we move the `TypeSystem` and the `value.proto` definitions from `Validation` into this repo, so that other renderers can also use them.

Doing so, we generalize `TypeSystem` so that it can be applied not only to Java, but all other target languages.
For that, we split `TypeSystem` into several components:
 - per se `TypeSystem` is a set of known Proto types;
 - `ValueConverter` converts `Value`s into code expressions in specific programming languages;
 - `TypeConvention` holds the knowledge on what language-specific definitions are generated for a given Proto type.

`ValueConverter` and `TypeConvention` are interfaces that have to be implemented for each specific language, while per se `TypeSystem` is language agnostic.

Also, now, `Renderer`s are part of the `Plugin` and should be supplied within the plugin in CLI and Gradle, not separately.

Also, now, `Renderer` has a type parameter for the target `Language`.


dmdashenkov:

compiler/src/main/kotlin/io/spine/protodata/backend/Values.kt
```diff
@@ -0,0 +1,158 @@
...
+ * A factory of `Value`s.
+ */
+public object Values {
+
+    private val NULL = value { nullValue = NULL_VALUE }
+
+    /**
+     * Converts the given message into a value.
+     *
+     *
```
armiol, line 65: This line is redundant.

compiler/src/main/proto/spine/protodata/value.proto
```diff
@@ -0,0 +1,140 @@
...
+
+        // Represents a structured value.
+        MessageValue message_value = 7;
+
+        EnumValue enum_value = 8;
+
+        // Represents a repeated `Value`.
+        ListValue list_value = 9;
+
+        // Represents a repeated `Value`.
```
armiol, line 62: Is it the same as previous field? Looks a copy-paste.

compiler/src/main/proto/spine/protodata/value.proto
```diff
@@ -0,0 +1,140 @@
...
+    // Can be used, for example, to distinguish between number values.
+    //
+    // If `list_value` is set, this is the type of the elements.
+    //
+    // If `map_value` is set, this is the type of the map values.
+    //
+    protodata.Type type = 21;
+}
+
+// `NullValue` is a singleton enumeration to represent the null value for the `Value` type union.
```
armiol, line 79: Let's also have back-ticks around `null`.

compiler/src/main/proto/spine/protodata/value.proto
```diff
@@ -0,0 +1,140 @@
...
+
+    repeated Entry value = 1;
+
+    // A single map entry.
+    message Entry {
+
+        // A map key.
+        //
+        // In Protobuf, a map key can only be of certain primitive types. See the official doc for
+        // more info: https://developers.google.com/protocol-buffers/docs/proto3#maps
```
armiol, line 133: I think we should put a period at the end. Even though it's a link.

armiol (changes requested):
@dmdashenkov I've completed the first round of review. I am planning to do another round once after I check this branch out locally.

dmdashenkov:
@armiol, PTAL again.

compiler/src/main/kotlin/io/spine/protodata/codegen/CodePrintable.kt
```diff
@@ -0,0 +1,40 @@
...
+ * A piece of code.
+ *
+ * This is an general interface for composable language-specific code-building elements.
+ */
+public interface CodePrintable {
+
+    /**
+     * Obtains the plain text representation of this code element.
+     */
+    public fun toCode(): String
```
armiol, line 39: In which case, maybe this is `text()`?

compiler/src/main/proto/spine/protodata/value.proto
```diff
@@ -0,0 +1,141 @@
...
+//
+// These definitions are inspired by the `google/protobuf/struct.proto`.
+//
+// Unlike types defined in `struct.proto`, these definitions keep strict typing at runtime. Values
+// are bound to their precise types.
+//
+// Also, unlike `struct.proto`, these definitions do not have special treatment when converting
+// to JSON. Consider using Google's types when JSON printing is required.
+
+// `Value` represents a dynamically typed value which can be either null, a bool, a number,
```
armiol, line 28:
```
`null`
```
?

compiler/src/main/proto/spine/protodata/value.proto
```diff
@@ -0,0 +1,141 @@
...
+// a string, a byte string, a recursive message value, an enum value, a list, or a map of values.
+//
+// Absence of any kind of value indicates an error.
+//
+message Value {
+    // The kind of value.
+    oneof kind {
+
+        // Represents a null value.
```
armiol, line 37: Same as above.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
```diff
@@ -0,0 +1,172 @@
...
+import io.spine.protodata.Value.KindCase.STRING_VALUE
+
+/**
+ * The type system of an application being built by ProtoData.
+ *
+ * The type system knows the association between the Protobuf types and the types of
+ * a target language.
+ *
+ * @param T the type of a language-specific type name
+ * @param V the type of a language-specific expression, i.e. the code that yields a value
```
alexander-yevsyukov, line 51: Now we separate this knowledge into two generic types, while they should correspond to each other. For example, I should not be able to match an `int` from C to `Integer` from Pascal. While currently these generics allow doing so.

The match between the types is and expression is a thing which should reside somewhere near `Language`.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * A [ValueConverter] for converting Protobuf values into language-specific code expressions.
+     */
+    protected abstract val valueConverter: ValueConverter<V>
+
+    /**
+     * Converts the given Protobuf type into a language-specific type name.
+     */
+    public fun typeNameToCode(type: Type): T {
```
alexander-yevsyukov, line 66: I don't think this method name fits the definition of the generic parameter `T`. `T` is a type name. We probably can obtain an expression out of `T`, so that it becomes `V` (while both are `CodePrintable`, which introduce additional confusion). But it's another story.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
```diff
@@ -0,0 +1,172 @@
...
+            LIST_VALUE -> toList(value)
+            MAP_VALUE -> toMap(value)
+            else -> throw IllegalArgumentException("Empty value")
+        }
+    }
+
+    /**
+     * A factory of language-specific code, that represents a Protobuf value.
+     */
+    public interface ValueConverter<V: CodePrintable> {
```
alexander-yevsyukov, line 114: Why do we need this interface nested? We have `Value` as a type. The converter to code in a certain language comes somewhere near. It does not relate directory to a type system. It relates to a programming language and expressions in it.

compiler/src/test/kotlin/io/spine/protodata/backend/ValuesSpec.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.Test
+import io.spine.protodata.test.postcard
+
+@DisplayName("`Values` should")
+class ValuesSpec {
+
+    @Test
+    fun `convert a simple message to a Value`() {
+        val v = Values.from(Time.currentTime())
+        v.messageValue.type.simpleName shouldBe "Timestamp"
```
alexander-yevsyukov, line 47: Can we have `with` or `run` around `v.messageValue`, please?

compiler/src/test/kotlin/io/spine/protodata/backend/ValuesSpec.kt
```diff
@@ -0,0 +1,77 @@
...
+        v.messageValue.fieldsMap["seconds"]!!.intValue shouldBeGreaterThan 0
+    }
+
+    @Test
+    fun `convert a repeated field`() {
+        val v = Values.from(fieldMask { paths.addAll(listOf("foo", "bar", "baz")) })
+        v.messageValue.type.qualifiedName() shouldBe FieldMask.getDescriptor().fullName
+        v.messageValue.fieldsMap shouldHaveKey "paths"
+        val list = v.messageValue.fieldsMap["paths"]!!.listValue
+        list.getValues(0).stringValue shouldBe "foo"
```
alexander-yevsyukov, line 58: Can we have `list.values[0]` somehow?

compiler/src/main/kotlin/io/spine/protodata/codegen/CodePrintable.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package io.spine.protodata.codegen
+
+/**
+ * A piece of code.
+ *
+ * This is an general interface for composable language-specific code-building elements.
+ */
+public interface CodePrintable {
```
alexander-yevsyukov, line 34: Why is it `Printable`? Why cannot it be just `Code`? Like `CodeBlock` in Poet, but shorter. I'm not sure such a generic word is a really good idea, but I'm not happy about the `Printable` suffix without a good reason.

We don't print source code on dead pieces of trees for ages. We used to do some 30 years ago. We've got bigger monitors with fast scrolling since then.

alexander-yevsyukov (changes requested):
Please see my comments.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
alexander-yevsyukov: If we convert a Proto type (name) to a language-specific type, I don't understand why we need a type system. The conversion is governed by code generation process for this kind of proto type and target programming language.

We don't need to gather all the types known to the application for this conversion. It's a matter of combination of two things: source Proto type and target programming language.

alexander-yevsyukov: I'm not against gathering all known types. We need this. But we don't do this for the conversion.

compiler/src/main/kotlin/io/spine/protodata/codegen/CodePrintable.kt
dmdashenkov: Let's leave this method as is for now. Changing the name will force us to change a lot of usages, since the hierarchy of which is new interface is a part, is heavily used already.

We can discuss if this name is fitting or not and, if so, do the change in a separate PR.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
dmdashenkov: We do need a type system to convert Proto names into Java named. That's the whole reason `TypeSystem` hierarchy exists.

Since we cannot always drag the whole info about the `.proto` file with us, and since we need that info to know how the generated type (e.g. Java class) will turn out, we need some kind of way to keep track of all the types.

For example, the `Type` in `fun typeNameToCode(type: Type)` knows nothing about outer classes and Java packages, yet it is a unique identifier for a Proto type.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
dmdashenkov: A `ValueConverter` is heavily intertwined with the `TypeSystem`. It needs the `TypeSystem` to know how to print a type's name in a language-specific manner.

The `TypeSystem` also serves as a public facade for the `ValueConverter`'s operations, so that the user has to deal with only one concept when developing code generation, not with two.

compiler/src/main/kotlin/io/spine/protodata/codegen/CodePrintable.kt
dmdashenkov: To print here means to convert into a string.

This is not really `Code` or `CodeBlock`. When I see `Code`, I assume that it is a snippet, one or more formatted, meaningful, and somewhat complete lines. But here, this is more of a building block for code, such as a type name, a method call, a conditional expression, an operator, etc.

dmdashenkov:

armiol (changes requested):
@dmdashenkov LGTM in general.

I second the comment of @alexander-yevsyukov, but I feel that's a matter of future PRs.

Requesting changes for now, since @alexander-yevsyukov is yet to review anyway.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
alexander-yevsyukov: I understand this is the consequence of the currently used decomposition. I asked the question because it looks as a design issue for me. Anyway, let's discuss it on Thursday.

compiler/src/main/kotlin/io/spine/protodata/codegen/CodePrintable.kt
alexander-yevsyukov: How about `CodeElement` then?

alexander-yevsyukov (changes requested):
Assuming the comments and suggestions I left, it's almost LGTM. Let's discuss in details on Thursday and merge it in soon.

@armiol, FYI.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
```diff
@@ -0,0 +1,172 @@
...
+import io.spine.protodata.Value.KindCase.NULL_VALUE
+import io.spine.protodata.Value.KindCase.STRING_VALUE
+
+/**
+ * An abstract base for type systems of an application being built by ProtoData.
+ *
+ * The type system knows the association between the Protobuf types and the types of
+ * a target language.
+ *
+ * @param T the type of a language-specific type name
```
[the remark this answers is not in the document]
alexander-yevsyukov: IDEA grammar checker also complains about "type of a". It insists on removing the indefinite article. Please have a look.

dmdashenkov:

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
```diff
@@ -48,13 +49,7 @@ private const val OF = "of"
...
  * Can be an expression, a reference to a variable, an identifier, etc.
  */
-public sealed interface JavaPrintable {
-
-    /**
-     * Prints this Java code.
-     */
-    public fun toCode(): String
-}
+public sealed interface JavaPrintable: CodeElement
```
alexander-yevsyukov, line 52: Maybe it's now `JavaElement`?

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * An abstract base for type systems of an application being built by ProtoData.
+ *
+ * The type system knows the association between the Protobuf types and the types of
+ * a target language.
+ *
+ * @param T the type of language-specific type name.
+ * @param V the type of language-specific expression, i.e. the code that yields a value.
+ */
+public abstract class BaseTypeSystem<T: CodeElement, V: CodeElement>
```
alexander-yevsyukov, line 42: Generic parameters don't do the work they supposed to do here.

I still don't understand why `BaseTypeSystem` deals with conversion at all. But even if it does, we should have at least something like `L: Language, T: CodeElement<L>, E: Expression<L>` coming as generic parameters.

Otherwise, it could be like `<KotlinElement, MarkdownText>` or something unbound like this.

compiler/src/main/kotlin/io/spine/protodata/codegen/BaseTypeSystem.kt
```diff
@@ -0,0 +1,79 @@
...
+            else -> unknownType(type)
+        }
+    }
+
+    /**
+     * Converts the given Protobuf primitive type into a language-specific type name.
+     *
+     * For example, `PrimitiveType.TYPE_INT64` is converted into `long` in Java.
+     */
+    protected abstract fun convertPrimitiveType(type: PrimitiveType): T
```
alexander-yevsyukov, line 64: It does not have to do anything with type system of an application at all. It's a matter of a programming language. 

I don't suggest extend `Language` with such things, but conversion should be next to `Language`, not within `BaseTypeSystem`.

compiler/src/main/kotlin/io/spine/protodata/codegen/CodeElement.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package io.spine.protodata.codegen
+
+/**
+ * A piece of code.
+ *
+ * This is an general interface for composable language-specific code-building elements.
+ */
+public interface CodeElement {
```
alexander-yevsyukov, line 34: Once again I suggest to have `CodeElement<L: Language>`.

alexander-yevsyukov (changes requested):
Please see my comments. Let's discuss further.

alexander-yevsyukov (changes requested):
Please see my comments.

dmdashenkov:
@alexander-yevsyukov, PTAL. There's still some work to do, mostly tests, but I'd like to hear your feedback on the general concepts first.

api/src/main/kotlin/io/spine/protodata/renderer/SourceFileSet.kt
```diff
@@ -316,3 +328,62 @@ private fun checkTarget(targetRoot: Path) {
...
+        val declaration = convention.declarationFor(name)
+        val path = declaration?.path
+        return path?.let { FileCreationWithPath(sources, path) }
+    }
+}
+
+/**
+ * Part of the fluent API for creating new source files.
+ */
+public class FileCreationWithPath(
```
alexander-yevsyukov, line 378: I think `FileLookup`, `FileCreation`, and `FileCreationWithPath` deserve a common interface which would group them semantically. These classes look like operations on `SourceFileSet`, and they all have `sources` as a property. 

I understand that they do not have a common API in terms of producing the output. And that could be the reason of not having a common interface for them. Still, grouping interface would help in documenting and discovering those operations.

api/src/main/kotlin/io/spine/protodata/type/TypeConventions.kt
```diff
@@ -0,0 +1,66 @@
...
+        conventions.asSequence()
+            .map { it.declarationFor(type) }
+            .filter { it != null }
+            .map { it!! }
+            .toSet()
+
+    /**
+     * Obtains a subset of these conventions for a specific language.
+     */
+    internal fun <LN: Language, TN : TypeNameElement<LN>>
```
alexander-yevsyukov, line 55: Please document the generic parameters. It's unclear why we have a different language type here. 

It looks as if languages could be nested as sets, and because of that we can obtain a subset of conventions.

api/src/main/kotlin/io/spine/protodata/type/TypeSystem.kt
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * A collection of known Protobuf types.
+ */
+public class TypeSystem
+
+/**
+ * Creates a new `TypeSystem` from the given definition files.
+ */
+public constructor(
```
alexander-yevsyukov, line 47: Why don't we want simple constructor here?

alexander-yevsyukov (commented):
Please see my comments.

compiler/src/main/kotlin/io/spine/protodata/backend/Pipeline.kt
```diff
@@ -57,18 +62,36 @@ import io.spine.server.under
...
+        return@lazy TypeConventions(cast)
+    }
+
+    public constructor(
+        plugins: List<Plugin>,
+        renderers: List<Renderer<*>>,
+        sources: List<SourceFileSet>,
+        request: CodeGeneratorRequest,
+        config: Configuration? = null
+    ) : this(plugins + ImplicitPluginWithRenderers(renderers), sources, request, config)
```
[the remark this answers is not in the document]
dmdashenkov: I've added comments on this in doc, as we've discussed.

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/JavaRejectionConvention.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.spine.protodata.TypeName
+import io.spine.protodata.type.GeneratedDeclaration
+import io.spine.protodata.type.TypeSystem
+import io.spine.tools.code.Java
+import io.spine.tools.code.Language
+
+/**
+ * A convention which governs Java Rejection-Throwable class declarations.
+ */
+public class JavaRejectionConvention(
```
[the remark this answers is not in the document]
dmdashenkov: Yes, as soon as rejections are generated via ProtoData.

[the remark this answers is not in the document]
alexander-yevsyukov: They are [already](https://github.com/SpineEventEngine/mc-java/pull/95).

alexander-yevsyukov (approved):
LGTM!

PR #153, opened 2023-07-23
Small fixes

alexander-yevsyukov:


armiol (approved):
@alexander-yevsyukov LGTM. 

@dmdashenkov please have a look, as you are one of the primary authors _and_ users of this API.

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/file/Options.kt
```diff
@@ -30,28 +30,29 @@ import com.google.protobuf.BoolValue
...
     this.name = "java_package"
     value = StringValue.of(name).pack()
-}.build()
+}
 
 /**
  * Obtains the option to set the [name] of the outer Java class.
  */
-public fun javaOuterClassName(name: String): Option = Option.newBuilder().apply {
+public fun javaOuterClassName(name: String): Option = option {
```
dmdashenkov, line 55: @alexander-yevsyukov, does your IDE work fine in here? Mine shows an error sometimes.

dmdashenkov (approved):
LGTM.

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/file/Options.kt
alexander-yevsyukov: @dmdashenkov, this is because deduplication of generated sources does not work in ProtoData. If you delete `build/generated/source/proto/kotlin` the error goes away.

Strangely, the deduplication works OK in `mc-java`. We tried to find out the difference between two projects with @armiol, but found nothing so far.

The only reason I can suggest now is that `GenerateProtoTask.excludeProtocOutput()` from `io.spine.protodata.gradle.plugin.Plugin.kt` is executed _earlier_ than all the tasks are configured in ProtoData Gradle project. And in `mc-java` timing is more favorable to us.

PR #154, opened 2023-07-28
Allow `Plugin` to tune `BoundedContextBuilder`

alexander-yevsyukov:
This PR adds a method `Plugin.extend(BoundedContextBuilders)`  which allows to tune the context being built the way a developer deem fit. The method is a way out of the current restricted protocol which allows a `Plugin` to expose `views()`, `viewRepositories()`, or `policies()`. 

In some causes we need `ProcessManager`s. E.g. if I want to remember a state of a proto file when it's discovered and do some work when it's going to be `FileExisted`, I do need a `ProcessManager`. Such a case could probably be handled via currently exposed API, but it's going to be awkward.

Other changes:
 * Latest `config` was applied.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/plugin/Plugin.kt
+++ b/api/src/main/kotlin/io/spine/protodata/plugin/Plugin.kt
@@ -26,6 +26,7 @@
 
 package io.spine.protodata.plugin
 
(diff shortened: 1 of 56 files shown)


armiol (approved):
@alexander-yevsyukov LGTM.

PR #156, opened 2023-08-05
Migrate to newer `ProcessManager` API

alexander-yevsyukov:
This PR migrates to newer `core-java` which recently introduced `ProcessManager.select()` method.

Also `Renderer.protoDataContext` `var` was made private in favor of new `@Internal` method `injectContext()`.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
+++ b/api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022, TeamDev. All rights reserved.
+ * Copyright 2023, TeamDev. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@
 
 package io.spine.protodata.renderer
 
+import io.spine.annotation.Internal
 import io.spine.base.EntityState
 import io.spine.protodata.config.ConfiguredQuerying
 import io.spine.server.BoundedContext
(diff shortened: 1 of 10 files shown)


api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
```diff
@@ -74,4 +75,21 @@ protected constructor(
...
+     *
+     * This method is `public` but is essentially `internal` to ProtoData SDK.
+     * It is annotated `@Internal` to prevent its accidental use by the users of the SDK.
+     *
+     * This method must be called by ProtoData only once.
+     * Repeat calls will result in `IllegalStateException`.
+     */
+    @Internal
+    public fun injectContext(protoDataContext: BoundedContext) {
+        check(!this::protoDataContext.isInitialized) {
```
dmdashenkov, line 90: Please use the `ContextAware` interface along with the check methods provided there.

dmdashenkov (changes requested):
Please see my comment.

api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
```diff
@@ -74,4 +75,28 @@ protected constructor(
...
+    public override fun registerWith(protoDataContext: BoundedContext) {
+        if (isRegistered) {
+            check(this.protoDataContext == protoDataContext) {
+                "Unable to register the renderer `$this` with" +
+                        " `${protoDataContext.name().value}`." +
+                        " The renderer is already registered with" +
+                        " `${this.protoDataContext.name().value}`."
+            }
+            return
+        }
```
dmdashenkov, line 95: Please use `checkNotRegistered()` instead.

dmdashenkov (approved):
LGTM with a single suggestion.

api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
alexander-yevsyukov: I thought of this, but `checkNotRegistered()` does not give enough diagnostics. It simply tells that an object is already registered. We need to tell a bit more in such a delicate case.

PR #158, opened 2023-08-10
Bump dependencies

alexander-yevsyukov:
This PR updates internal and external dependencies, with the primary focus on getting newer version of Logging library, which fixes the accessor of the `WithLogging.logger` property.

Other changes:
  * `NoOp` implementation of `SourceAtPoint` no longer implements deprecated `Logging` interface.


buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
```diff
@@ -33,7 +33,7 @@ package io.spine.internal.dependency
  */
 @Suppress("unused", "ConstPropertyName")
 object ProtoData {
-    const val version = "0.9.9"
+    const val version = "0.9.0"
```
dmdashenkov, line 36: Is this downgrade an intentional one?

alexander-yevsyukov: I was trying to overcome the `launchProtoData` issue caused with the `logging` API misstep, and rolled back the version of ProtoData. But it does not have effect because it's McJava, which is used for codegen. So, the rollback of its version helped instead. Still, I'll put it to latest `0.9.10`, just for consistency.

dmdashenkov (approved):

PR #159, opened 2023-08-17
Make standard renderers self-sufficient plugins

dmdashenkov:
In this PR we make case-specifc renderers declared in ProtoData into plugins, so that end users may stop applying them as separate renderers and use plugins instead.

For a smooth migration with no additional headache around renaming classes, make the renderers implement the `Plugin` interface. For example, `io.spine.protodata.codegen.java.annotation.GeneratedTypeAnnotation` is now both a renderer and a plugin, GOF-Adapter-style.

Also, the method `Renderer.typeConventions()` is renamed into `knownTypeConventions()` to make the Adapter trick work by avoiding name clashes.

In the future, those classes should be split to have a separate `Renderer`, as soon as the ability to pass standalone renderers is removed.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
+++ b/api/src/main/kotlin/io/spine/protodata/renderer/Renderer.kt
@@ -81,7 +81,7 @@ protected constructor(
      *
      * @param N the type of the name element for the associated language
      */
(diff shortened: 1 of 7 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #160, opened 2023-08-22
JVM static methods

dmdashenkov:
In this PR we make some new Kotlin API more compatible with Java by introducing `@JvmStaticMethod` annotations.

We also update Spine dependencies, logging in particular.


dmdashenkov:
@alexander-yevsyukov, FYI.

alexander-yevsyukov:
@dmdashenkov, the ”mystery” was implementation of the logging context, which previously was our own. Recently, `core-java` got dependency on gRPC-based logging context. Having two implementations of the logging context implementation (which are `AutoService`-es`) cased Logging to fallback to default (no-op) implementation. That's why `LogLevelMap` stopped working.

I alo disabled the test which assumes that `WARNING` is the default level. When we have JUL as the logging backend (which is our case), the default level is `INFO`, and it takes some tricks to make it stricter than `INFO`. If this is really needed, it can be done.

@yevhenii-nadtochii, @armiol, FYI.

dmdashenkov:

dmdashenkov:
@alexander-yevsyukov, I removed the test you mentioned completely. It was added previously in this branch while trying to find the root cause of the logging issue.

armiol (commented):
@dmdashenkov as for the changes themselves, LGTM. 

However, it would be nice to have some usages in pure Java tests, so that we are sure the methods marked `@JvmStatic` are accessible in a way we intend them to be.

compiler/src/test/java/io/spine/protodata/backend/MoreValuesSpec.java
```diff
@@ -0,0 +1,54 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata.backend;import io.spine.protodata.backend.Values;
```
alexander-yevsyukov, line 27: Please have the `import` on a separate line.

alexander-yevsyukov (approved):
LGTM, with one formatting request.

PR #168, opened 2023-09-01
Pull important types out. Expose extensions functions

alexander-yevsyukov:
This PR:
  * Pulls `ClassName` and `JavaElement` to the level of their package. 
  * Methods of `ClassName` when it was under `Expression.kt` were transformed into extension functions.
  * Makes most of the extensions declared in `CodeLayout.kt` `public` so that we can move `JavaRejectionRejectionConvention` to `mc-java` sometime soon.
  * `ClassName` got additional constructor with `vararg` parameter so that we don't have to do `listOf()` all the time we need to pass one or a couple of strings.

### DIFF
--- a/codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
+++ b/codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022, TeamDev. All rights reserved.
+ * Copyright 2023, TeamDev. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/Expression.kt
```diff
@@ -129,128 +114,46 @@ public class LiteralBytes(bytes: ByteString) : Expression(
...
-    }
-    val names = mutableListOf<String>()
-    var cls: Class<*>? = this
-    do {
-        names.add(cls!!.simpleName)
-        cls = cls.declaringClass
-    } while (cls != null)
-    return names.reversed()
-}
+public fun ClassName.enumValue(number: Int): MethodCall =
```
armiol, line 155: Let's discuss moving public `ClassName` extensions to `ClassName.kt`.

armiol (changes requested):
@alexander-yevsyukov LGTM except for two things:

1. See my comment on joining the forces of `ClassName` and its extensions in scope of a single `ClassName.kt` file.
2. The PR description seems to end abruptly.

armiol (approved):

PR #171, opened 2023-09-04
Naming and doc improvements

### LINKED ISSUE #163: Improve the name of `TypeSystem.from(Querying)`
The method `TypeSystem.from()` is typically used by `Renderer`s. So the usage looks like this:

```kotlin
val typeSystem = TypeSystem.from(this)
```

It does not read well because we do not create a `TypeSystem` from rendering. We create it "for" rendering. Obviously, we cannot use `for` as it is a reserved word. But `from` isn't a good replacement because it's confusing.

Let's have a more meaningful name. How about having it like this:

```kotlin
val typeSystem = TypeSystem.serving(this)
```

### LINKED ISSUE #164: Describe nullability for `TypeConvention.declarationFor()`
Here's how the method is declared:

```kotlin
    /**
     * Given a Protobuf type name, obtains the primary declaration generated from this Proto type.
     */
    public fun declarationFor(name: TypeName): GeneratedDeclaration<L, T>?
```

In which cases the returned declaration is `null`?

### LINKED ISSUE #165: `MessageTypeConvention` needs either a new name or more docs, or both
The class has `declarationFor()` method implemented as shown below:

```kotlin
    override fun declarationFor(name: TypeName): GeneratedDeclaration<Java, ClassName> {
        val file = typeSystem.findMessageOrEnum(name)?.second
        check(file != null) { "Unknown type `${name.typeUrl}`." }
        val cls = name.javaClassName(declaredIn = file)
        return GeneratedDeclaration(cls, cls.javaFile)
    }
```

Assuming, we call `findMessageOrEnum()` this convention should suite not only message types, but enums too. 

The problem with the current name is that it does not suggest the fact. I faced it when porting rejection generation code to newer `TypeSystem` coming from ProtoData.

dmdashenkov:
In this PR we make small improvements in the ProtoData's API:
 - `TypeSystem.from(..)` renamed to `TypeSystem.serving(..)` — fixes #163;
 - doc for `TypeConvention.declarationFor(..)` is improved to account for  `null` return values — fixes #164;
 - `MessageTypeConvention` renamed to `JavaImplConvention` — fixes #165.

### DIFF
--- a/codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/JavaRejectionConvention.kt
+++ b/codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/JavaRejectionConvention.kt
@@ -35,6 +35,9 @@ import io.spine.tools.code.Language
 
 /**
  * A convention which governs Java Rejection-Throwable class declarations.
+ *
+ * The convention only defines a declaration for rejection message types. Any other types are
+ * undefined and thus result in the [declarationFor] method returning `null`.
  */
 public class JavaRejectionConvention(
     typeSystem: TypeSystem
--- a/api/src/main/kotlin/io/spine/protodata/type/Declaration.kt
+++ b/api/src/main/kotlin/io/spine/protodata/type/Declaration.kt
@@ -28,16 +28,25 @@ package io.spine.protodata.type
 
 import io.spine.tools.code.Language
 import java.nio.file.Path
+import kotlin.DeprecationLevel.ERROR
 
 /**
- * A generated type declaration.
+ * A type declaration that can be generated from a Protobuf type.
  *
  * @property name the name of the generated type.
  * @property path the path to the file where the declaration is placed.
  *
+ * @param L the type of the target language.
  * @param T the language-specific type of the type name element.
  */
-public data class GeneratedDeclaration<L: Language, T : TypeNameElement<L>>(
+public data class Declaration<L: Language, T : TypeNameElement<L>>(
     public val name: T,
     public val path: Path
 )
+
+@Deprecated(
+    "Use `Declaration` instead.",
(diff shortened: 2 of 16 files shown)


dmdashenkov:

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/JavaRejectionConvention.kt
```diff
@@ -35,6 +35,9 @@ import io.spine.tools.code.Language
 
 /**
  * A convention which governs Java Rejection-Throwable class declarations.
+ *
+ * The convention only defines a declaration for rejection message types. Any other types are
+ * undefined and thus result in the [declarationFor] method returning `null`.
  */
 public class JavaRejectionConvention(
```
alexander-yevsyukov, line 42: What if we rename this class to `RejectionThrowableConvention`? We have already `java` in the package.

Also, our rejections are two-headed. One "head" is a message, another extends `RejectionThrowable`. Having a more precise name for this class would help to understand its purpose a bit faster.

I am going to grab this class and move to `mc-java` soon. I tried it but couldn't do it because some extension functions were `internal`.

alexander-yevsyukov: Also, I had a comment regarding the name `GeneratedDeclaration`, especially in the context of this class.

The convention for `RejectionThrowable` codegen does not refer to already generated class. We are going to generate this class having this convention.

Therefore, I think that the `Generated` prefix is redundant because it refers to a stage of source code processing, which does not apply in all cases.

dmdashenkov:

alexander-yevsyukov (approved):
LGTM

PR #173, opened 2023-09-08
Fix filtering

dmdashenkov:


dmdashenkov:
@alexander-yevsyukov, PTAL. @armiol, PTAL.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Extension.kt
```diff
@@ -81,12 +83,64 @@ public class Extension(internal val project: Project): CodegenSettings {
...
+    override val paths: Multimap<String, SourcePaths> = HashMultimap.create()
+
+    /**
+     * Obtains the source configured paths.
+     *
+     * If the deprecated [subDirs] and [targetBaseDir] are used, constructs [SourcePaths] instances
+     * from the present data for backward compatibility. However, in this compatibility mode,
+     * only Java and Kotlin source file sets can be constructed.
+     */
+    internal fun pathsOrCompat(sourceSet: SourceSet): Set<SourcePaths> {
```
armiol, line 95: The problem here is that we need to remember which version (this one or `paths`) to use. And there is no clear sign on when to stop using one and start using another.

Maybe, we could have a "compatibility" mode turned on always. And for testing (or whatever we need "strictly new API" mode for) have another method, which would ignore the deprecated properties?

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Extension.kt
```diff
@@ -81,12 +83,64 @@ public class Extension(internal val project: Project): CodegenSettings {
...
+        if (!paths.isEmpty) {
+            return paths[sourceSet.name].toSet()
+        }
+        val srcRoots = sourceDirs(sourceSet).get()
+        val targetRoots = targetDirs(sourceSet).get()
+        return srcRoots.zip(targetRoots)
+            .map { (src, target) ->
+                val pathSuffix = src.asFile.toPath().name
+                val lang = if (pathSuffix == "kotlin") Kotlin else Java
```
armiol, line 104: Instead of using a literal, maybe we could use `knownLanguages`, which we have previously created exactly to associate `Kotlin` with `"kotlin"`?

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Extension.kt
```diff
@@ -81,12 +83,64 @@ public class Extension(internal val project: Project): CodegenSettings {
...
+
+    public companion object {
+
+        /**
+         * Default subdirectories expected by ProtoData under a generated source set.
+         *
+         * @see subDirs
+         */
+        public val defaultSubdirectories: List<String> = listOf(
+            "java",
```
armiol, line 127: With all these words constantly repeating, it again strikes me on why we don't re-use the definitions, say, from either languages, or generators (or labels), that we know exist.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Extension.kt
```diff
@@ -81,12 +83,64 @@ public class Extension(internal val project: Project): CodegenSettings {
...
+            "kotlin",
+            "grpc",
+            "js",
+            "dart",
+            "spine",
```
armiol, line 132: I think we don't generate anything into "spine" for a while already.

armiol (changes requested):

api/src/main/kotlin/io/spine/protodata/renderer/SourceGeneratorName.kt
```diff
@@ -0,0 +1,67 @@
...
+ * the `DefaultGenerator` is only defined for those languages that are supported, such as Java, JS,
+ * C++, etc. For other languages, as well as for other code generation scenarios,
+ * see [CustomGenerator].
+ */
+public object DefaultGenerator : SourceGeneratorName
+
+/**
+ * A name of a custom source code generator.
+ *
+ * May represent a Protobuf compiler plugin, or any other code generator.
```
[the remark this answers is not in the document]
dmdashenkov: "Default" simply references the case most users will likely face most of the time. And it is the default way in which Protoc generated code, as opposed to custom generators/Protoc plugins.

I've cleared up the doc's wording a bit so that there is no confusion with the labels.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Extension.kt
dmdashenkov: I've added a note on this in doc. Essentially, this needs to exist until `mc-java` is updated. Potentially, until the very next PR. Then, only `paths` need to be used.

gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Extension.kt
dmdashenkov: Some here. This list will die off once the old `subdirs` API is removed.

alexander-yevsyukov (changes requested):
I don't understand why we need extra settings for ProtoData while we know what Protobuf and its Gradle Plugin have for languages and generators, like gRPC.

It's either a plugin or built-in in Protobuf codegen terms. The set of supported builtins is known. There are also known plugins like gRPC.

We can get this information from there and transform into `Language` and `Generator` terms.

alexander-yevsyukov:
> We can get this information from there and transform into `Language` and `Generator` terms.

When I wrote this I meant we don't need yet another term "label".

armiol (changes requested):
@dmdashenkov basically, the gist of this PR is as follows. 

We could have discussed the approach beforehand, or deal with a number of newly introduced language terms in scope of PR. The latter is longer, but here we are, and there is not much time left to make things conclude with a resolution leaving all parties happy. 

I suggest we gather tomorrow on a call with @alexander-yevsyukov and discuss these changes again, instead of bouncing thing back and forth in a PR.

dmdashenkov:
@armiol, @alexander-yevsyukov, PTAL at the current state of this changeset. I've reflected the main points of our discussion in the PR header and changed some of the naming accordingly.

PR #174, opened 2023-09-13
Better annotation code

alexander-yevsyukov:
This PR improves code generation of Java class annotations. In particular:
  * The `TypeAnnotation` abstract class no longer implements `Plugin`. It's not a really good idea to have such a mixture in an abstract class.
  * Instead of mixing `Renderer` and `Plugin` in abstract classes, `AbstractPlugin` class and `TypeAnnotationPlugin` classes were introduced. 
  * The descendants of `TypeAnnotation` now has  nested `Plugin` classes which makes them reusable `Renderer`s in other plugins.


dmdashenkov (approved):
LGTM.

PR #183, opened 2023-10-19
Improve type system injection

alexander-yevsyukov:
This PR simplifies the way an instance of `TypeSystem` is injected into a `Renderer`:
  * The method `injectTypeSystem()` was made `public` and annotated `@Internal`. It's way simpler than having "gymnastics" with constructors down the `Renderer` inheritance hierarchy.
  * The constructor accepting `TypeSystem` as a second parameter was removed.
  * The contract for `injectTypeSystem()` was made stricter to allow passing only one instance of `TypeSystem`, or repeated calls with the same instance.

### DIFF
--- a/api/src/test/kotlin/io/spine/protodata/renderer/RendererSpec.kt
+++ b/api/src/test/kotlin/io/spine/protodata/renderer/RendererSpec.kt
@@ -71,14 +71,24 @@ internal class RendererSpec {
 
     companion object {
 
+        /**
+         * An instance of [CodeGenerationContext] to be used for
+         * [creating][createTypeSystem] a [typeSystem].
+         *
+         * We do not pass any custom entity classes or repositories to the context builder,
+         * as we do not need them in tests.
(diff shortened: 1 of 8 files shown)


api/src/test/kotlin/io/spine/protodata/renderer/RendererSpec.kt
```diff
@@ -0,0 +1,94 @@
...
+    @Test
+    fun `prevent injecting another type system`() {
+        assertThrows<IllegalStateException> {
+            renderer.injectTypeSystem(createTypeSystem())
+        }
+    }
+
+    companion object {
+
+        private val context = CodeGenerationContext.builder().build()
```
armiol, line 74: This "default" instance looks weird.

Could you please add a comment on why we build it like that here?

armiol (approved):
@alexander-yevsyukov LGTM with a comment to consider.

PR #185, opened 2023-10-26
Introduce `CodegenContext` interface

alexander-yevsyukov:
Before this PR the bounded contexts involved with the code generation had fixed names. In some cases the names of these contexts were used as IDs of the event producers. This, in turn, could led to dispatching errors if more than one `Pipeline` were started e.g., in parallel or one after another (without closing the previously used contexts before starting the 2nd `Pipeline`).

In order to overcome this limitation, this PR introduces the `Pipeline.id` property, which is used as a suffix for the context names of the pipeline.  By default the ID is a generated securely random value (see `SecureRandomString` object). It also can be set as a value passed to the `Pipeline` constructor.

### `io.spine.protodata.CodegenContext` interface
The interface was introduced for being able to propagate an instance of `TypeSystem` to user-defined `Plugin`s. The interface also expose `insertionPointsContext` which can be of interested for author of renderers. 

### Other changes
 * Kotlin proto DSL is used for generating compiler events.
 * Conventions for property names were documented (https://bit.ly/kotlin-prop-names) and used for suppression of `"ConstPropertyName"` warning.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/CodegenContext.kt
+++ b/api/src/main/kotlin/io/spine/protodata/CodegenContext.kt
@@ -24,25 +24,33 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.protodata.type;
+package io.spine.protodata
 
-import io.spine.base.EntityState;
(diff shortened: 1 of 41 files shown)


alexander-yevsyukov:
@armiol, here's the PR with the recent changes in codegen tests.

armiol (approved):

PR #188, opened 2023-11-13
Improve type names and extensions

alexander-yevsyukov:
This PR improves the names of proto types used in a domain model of code generation. Some types and their properties were renamed to better reflect the nature of data they represent.

## Renamed types and properties
 * `File` -> `ProtoFileHeader` because it carries options, the package and syntax info. It's more than a file.
 * Correspondingly, `File.path` became `ProtoFileHeader.file`.

The previous renamings allowed us to do these renamings:
 * `FilePath.value` -> `File.path`. It's a proto replacement for `kotlin.File`.

In other types properties that used to be `path` or `file_path` became `file`. 
Properties that used to be `file` (referring to `ProtobufSourceFile`) became `source.

## Added extension properties

`Field` got the following extension properties:
 * `isMessage` — it reads better than `Field.type.hasMessage()`, which we get generated from proto type.
 * `isMap`
 * `isList` became property instead of `isList()` function. This is to highlight the fact that a field cannot change its nature once created.
 * `isRepeated` — similarly to `isList` became a property.
 * `isPartOfOneof`  — similarly to `isList` became a property.
 * `qualifiedName` — returns the property name including the qualified name of the type which declares it.

`Type` got these extension properties:
 * `isMessage`
 * `isAny`

## `FileAware` and `OptionAware` interface were introduced
Events that are associated with a `File` became `FileAware`.
Events that are associated with an `Option` became `OptionAware`.

## Other changes
 * `Renderer` got inline function `select()` for idiomatic calls from Kotlin.
 * Removed `skipValidation()` in builds because it's no longer needed.
 * `SecureRandomString` made public for usage in Validation.
 * Constants related to ProtoData configuration were added to `Names` object for usage in McJava.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
+++ b/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
@@ -118,7 +118,7 @@ public val Field.isPartOfOneof: Boolean
     get() = hasOneofName()
 
 /**
- * Obtains the name of the field which includes a qualified name of the type which declares it.
+ * The field name containing a qualified name of the declaring type.
  */
 public val Field.qualifiedName: String
     get() = "${declaringType.qualifiedName}.${name.value}"
--- a/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
+++ b/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
@@ -48,7 +48,7 @@ import io.spine.protodata.CallCardinality.UNARY
  * Obtains a name of this Protobuf file without the extension.
  */
 public fun ProtoFileHeader.nameWithoutExtension(): String {
-    val name = file.value.split("/").last()
+    val name = file.path.split("/").last()
     val index = name.indexOf(".")
     return if (index > 0) {
         name.substring(0, index)
@@ -190,7 +190,7 @@ public fun FieldDescriptor.name(): FieldName = fieldName { value = name }
 /**
  * Obtains the relative path to this file as a [FilePath].
  */
-public fun FileDescriptor.file(): FilePath = filePath { value = name }
+public fun FileDescriptor.file(): File = file { path = name }
 
 /**
  * Obtains the name of this service as a [ServiceName].
--- a/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
+++ b/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
@@ -47,7 +47,7 @@ import io.spine.protodata.CallCardinality.UNARY
 /**
  * Obtains a name of this Protobuf file without the extension.
  */
-public fun File.nameWithoutExtension(): String {
+public fun ProtoFileHeader.nameWithoutExtension(): String {
     val name = path.value.split("/").last()
     val index = name.indexOf(".")
     return if (index > 0) {
--- a/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
+++ b/api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
@@ -60,19 +60,27 @@ public fun File.nameWithoutExtension(): String {
 /**
  * Obtains the package and the name of the type.
  */
-public fun MessageType.qualifiedName(): String = name.qualifiedName()
+public val MessageType.qualifiedName: String
+    get() = name.qualifiedName
 
 /**
  * Obtains the fully qualified name from this `TypeName`.
  */
-public fun TypeNameOrBuilder.qualifiedName(): String {
-    val names = buildList<String> {
-        add(packageName)
-        addAll(nestingTypeNameList)
-        add(simpleName)
+public val TypeNameOrBuilder.qualifiedName: String
+    get() {
+        val names = buildList<String> {
+            add(packageName)
+            addAll(nestingTypeNameList)
+            add(simpleName)
+        }
+        return names.filter { it.isNotEmpty() }.joinToString(separator = ".")
     }
-    return names.filter { it.isNotEmpty() }.joinToString(separator = ".")
-}
+
+/**
+ * Tells if this field is a Protobuf message.
+ */
+public val Field.isMessage: Boolean
+    get() = type.hasMessage()
 
 /**
  * Shows if this field is a `map`.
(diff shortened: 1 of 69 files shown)


compiler/src/main/kotlin/io/spine/protodata/backend/event/OptionsCompilerEvents.kt
```diff
@@ -55,10 +55,10 @@ internal suspend fun SequenceScope<EventMessage>.produceOptionEvents(
 }
 
 /**
- * Parses the given `options` message into a list of [Option]s.
+ * Parses the this `options` message into a list of [Option]s.
```
yevhenii-nadtochii, line 58:
```suggestion
 * Parses this `options` message into a list of [Option]s.
```

gradle-api/src/main/kotlin/io/spine/protodata/gradle/Names.kt
```diff
@@ -45,4 +50,15 @@ public object Names {
...
     public const val EXTENSION_NAME: String = "protoData"
+
+    /**
+     * The name of the Gradle Configuration created by ProtoData Gradle plugin for holding
+     * user-defined classpath.
+     */
+    public const val USER_CLASSPATH_CONFIGURATION: String = "protoData"
+
+    /**
+     * The name of the configuration containing to ProtoData `fat-cli` artifact.
```
yevhenii-nadtochii, line 61:
```suggestion
     * The name of the configuration containing ProtoData `fat-cli` artifact.
```

test-env/src/main/kotlin/io/spine/protodata/test/InternalAccessRenderer.kt
```diff
@@ -41,7 +41,7 @@ public class InternalAccessRenderer : SoloRenderer<Java>(Java) {
     override fun render(sources: SourceFileSet) {
         val internalTypes = select<InternalType>().all()
         internalTypes.forEach { internalType ->
-            val path = internalType.name.qualifiedName().replace('.', File.separatorChar)
+            val path = internalType.name.qualifiedName.replace('.', File.separatorChar)
```
yevhenii-nadtochii, line 44:
```suggestion
            val path = internalType.qualifiedName.replace('.', File.separatorChar)
```
```suggestion
            val path = internalType.name.qualified.replace('.', File.separatorChar)
```

I don't know whether it is possible, but now it looks a bit wordy. But not critical.

yevhenii-nadtochii (approved):
LGTM with comments to enhance docs.

api/src/main/proto/spine/protodata/insertion.proto
```diff
@@ -61,7 +61,7 @@ message InsertedPoints {
     option (entity).kind = VIEW;
 
     // The relative path to the code file.
```
armiol, line 63: I would append the notion of possible "relativeness" to the definition of `File`.

codegen/java/src/main/kotlin/io/spine/protodata/codegen/java/CodeLayout.kt
```diff
@@ -105,9 +105,9 @@ internal fun composeJavaClassName(
 }
 
 /**
- * Obtains a fully-qualified Java class, generated for the Protobuf type with this name.
+ * Obtains a fully qualified Java class, generated for the Protobuf type with this name.
  */
-internal fun TypeName.javaClassName(declaredIn: File): ClassName =
+internal fun TypeName.javaClassName(declaredIn: ProtoFileHeader): ClassName =
```
armiol, line 110: Let's discuss this signature and those below. They read weird. Maybe, it's not `declaredIn` anymore?

tests/protodata-extension/src/main/proto/spine/protodata/test/uuid_type.proto
```diff
@@ -20,5 +20,5 @@ message UuidType {
     protodata.TypeName name = 1;
 
     // The file where the type is declared.
-    protodata.FilePath declared_in = 2;
+    protodata.File declared_in = 2;
```
armiol, line 23: See? Here `declared_in` still makes sense, because it's about `File`.

armiol (changes requested):
@alexander-yevsyukov Please see my comments.

Let's discuss the Proto file header as a type, and its usages within this changeset (see my comments).

api/src/main/kotlin/io/spine/protodata/AstExtensions.kt
```diff
@@ -47,8 +47,8 @@ import io.spine.protodata.CallCardinality.UNARY
 /**
  * Obtains a name of this Protobuf file without the extension.
```
armiol, line 48: Let's also modify this description, since it's a header's extension now.

api/src/main/proto/spine/protodata/events.proto
```diff
@@ -38,252 +38,330 @@ option java_multiple_files = true;
...
+
 // Emitted when processing of a file begins.
 message FileEntered {
 
-    FilePath path = 1;
+    // The file processing of which is started.
+    File file = 1;
 
-    File file = 2;
+    // The header of the file.
```
armiol, line 49: Let's append this notion of the equality of `file` and `header.file`, we do in another file in this changeset.

api/src/main/proto/spine/protodata/source.proto
```diff
@@ -46,11 +46,12 @@ message ProtobufSourceFile {
...
     // The relative path to the file.
     //
-    // Must be the same as `file.path`.
+    // Must be the same as `header.file`.
     //
-    FilePath file_path = 1;
+    File file = 1;
 
-    File file = 2;
+    // A header of the Protobuf file.
```
armiol, line 53: Let's append this notion of the equality of `file` and `header.file`, we do in another file in this changeset.

codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/JavaCodegenSpec.kt
```diff
@@ -45,16 +45,16 @@ import org.junit.jupiter.api.Test
...
-    @Nested
-    inner class `Obtain Java class name from` {
+    @Nested inner class
+    `Obtain Java class name from` {
 
         @Test
         fun `top-level message`() {
             val typeName = "Anvil"
             val type = messageType(typeName)
             val file = protoMultipleFiles()
```
armiol, line 55: Here and below, `file` variables should probably be `header`.

And maybe the method `protoMultipleFiles()` should be renamed as well.

armiol (changes requested):
@alexander-yevsyukov there are a couple of more comments, that seem apparent after `accordingTo` is introduced.

armiol (approved):

PR #192, opened 2023-12-25
Improve type annotations

alexander-yevsyukov:
This PR extends `psi-java` artifact recently introduced in `tool-base` for parsing Java files:
 * `PsiJavaParser` facade was introduced to simplify the creation of a PSI `Project`, and an instance of `Parser` associated with it.
 * `BeforePrimaryDeclaration` insertion point was migrated to use `PsiJavaFile`.

Also this PR improves `TypeAnnotation` allowing it to work with nested types:
 * A hierarchy of Java type names was introduced, `EnumName` and `ClassName` being the leaves of the hierarchy.
 * `BeforeNestedTypeDeclaration` insertion point was introduced.
 * `TypeAnnotation` also got the `subject` property to simplify the usage when the type to be annotated is already known.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -73,6 +73,10 @@ allprojects {
     version = extra["protoDataVersion"]!!
 
     repositories.standardToSpineSdk()
+    repositories {
+        intellijReleases
+        jetBrainsCacheRedirector
+    }
 
     configurations.all {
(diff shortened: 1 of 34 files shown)


armiol (approved):
@alexander-yevsyukov LGTM.

PR #193, opened 2023-12-27
Allow label-less `InsertionPoint` to be used in `SourceAtLine.add()`

alexander-yevsyukov:
This PR improves the way `SourceAtLine.add()` method works with insertion points.

Previously, the method relied an an code label an insertion point printer added to the source code. This approach does not work for insertion points like `BeforeNestedTypeDeclaration`, which rely on PSI.

New implementation of the `add()` method relies on `InsertionPoint.locate()` method, and is compatible with insertion points that do not have labels.

Other notable changes:
  * `InsertionPointPrinter` now encourages passing only points with non-empty label. Its constructor now accepts `Iterable<InsertionPoint>` and checks for non-empty labels. It is still possible to have such points via overloading `supportedInsertionPoints()` method, which became deprecated. The updated printing implementation simply filters such points out.
  * Added integration tests for `BeforeNestedTypeDeclaration` and `BeforePrimaryDeclaration`.
  * `ImplicitPluginWithRenderers` class became `public` to encourage usage for integration tests of custom renderers.
  * Test constructor became public in `Pipeline`.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
+++ b/api/src/main/kotlin/io/spine/protodata/renderer/InsertionPoint.kt
@@ -43,6 +43,17 @@ public interface InsertionPoint : CoordinatesFactory, WithLogging {
 
     /**
      * The name of this insertion point.
+     *
+     * The name of the insertion point is used to formally identify the place
+     * where the code should be inserted. It could be a place in the code marked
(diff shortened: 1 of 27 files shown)


armiol (approved):
@alexander-yevsyukov LGTM with a minor comment to address.

Issue #196, opened 2024-01-06
Gradle plugin should add `grpc` directory to the `java` source set


alexander-yevsyukov:
When proto files have some services, gRPC codegen generates the code as expected. But the `generated/<sourceSet>/grpc/` directory is not added the corresponding source set in the Gradle project.

Here's what it takes to compensate for this missing feature:

```kotlin
    sourceSets {
        testFixtures {
            java.srcDirs("generated/testFixtures/grpc")
        }
    }
```

ProtoData should do it in its Gradle plugin to reduce the manual labor required from the programmer.

alexander-yevsyukov:
Addressed by $197.

alexander-yevsyukov:
Hopefully, #199 addresses this issue for good.

PR #197, opened 2024-01-11
Add generated gRPC directory

alexander-yevsyukov:
This PR address #196, adding `grpc` directory to the `java` source set, if gRPC codegen is configured in a project.

Other notable changes:
 * The filtering for avoiding duplicated source code was improved.

### DIFF
--- a/gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Plugin.kt
+++ b/gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Plugin.kt
@@ -44,9 +44,12 @@ import io.spine.protodata.gradle.Names.USER_CLASSPATH_CONFIGURATION
 import io.spine.protodata.gradle.ProtocPluginArtifact
 import io.spine.tools.code.manifest.Version
 import io.spine.tools.gradle.project.sourceSets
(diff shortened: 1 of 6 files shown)


gradle-plugin/src/main/kotlin/io/spine/protodata/gradle/plugin/Plugin.kt
```diff
@@ -281,67 +284,133 @@ private fun Project.configureProtoTask(task: GenerateProtoTask, ext: Extension)
...
- * The current Protobuf support of Kotlin is based on Java codegen. Therefore,
- * it's likely that Java would be enabled in the project for Kotlin proto
- * code to be generated. Though, it may change someday, and Kotlin support for Protobuf would be
+ * @return `true` if `java` plugin is installed, `false` otherwise.
+ */
+private fun Project.hasJava(): Boolean =
+    pluginManager.hasPlugin("java")
+
+/**
+ * Tells if this project may deal with Kotlin code.
```
armiol, line 295: I would say "can" (like below), or "is able". 

"May" is not about a capability, which we characterize here, it's rather about a permission, or a probability.

armiol (approved):
@alexander-yevsyukov, LGTM. See my comment, though.

PR #199, opened 2024-01-14
Add `grpc` directory unconditionally

alexander-yevsyukov:
This PR changes the way a `generated/.../grpc` subdirectory was added to a source set. 

Now we do not check if `grpc` plugin is added to `protoc`. Plugins for a proto task may not be available at the time we previously checked. It is safe to do so because adding source set directories to a source set does not require that a directory must exist.


armiol (approved):
LGTM. Let's hope this `srcDir` behaviour remains the same in future Gradle versions (a smiling-crying face here).

PR #204, opened 2024-01-18
File-only settings

alexander-yevsyukov:
This PR makes ProtoData use only file-based settings that are stored under a common directory. 

The directory is specified in a new command-line parameter. Command-line parameters for configuration file or configuration value (along with its format) were eliminated. Correspondingly,  types and events that were related to non-file settings were removed as well.

`SettingsDirectory` class was introduced to handle reading and writing setting files. The instance of the class is now passed to `Pipeline` instead of `Config`.

### Other notable changes
  * Added inline extension function that allows to `add<MyViewClass>()` to `BoundedContextBuilder`.
  * `Format.PLAIN` got `txt` extension so that we can read/write such settings with files.
  * `DokkaTask`s are not executed only if they are in the public publishing task graph. This improves the build performance especially in cases when Dokka could not download some resources form internet.


api/src/main/kotlin/io/spine/protodata/settings/DiscoveredSettings.kt
armiol, line 1: 🎆

api/src/main/kotlin/io/spine/protodata/settings/SettingsDirectory.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Writes settings file for the given consumer.
+     *
+     * @param consumerId the ID of the consumer to write settings for.
+     * @param format the format of the settings file.
+     * @param content the content of the settings file.
+     */
+    public fun write(consumerId: String, format: Format, content: ByteArray) {
```
armiol, line 84: Eventually, we should not have this `consumerId` as `String`. It's because right now the first parameter of `write(String, Format, ByteArray)` can be easily mis-interpreted as the content to write to the file.

armiol (approved):
@alexander-yevsyukov LGTM with a non-critical comment to discuss some time later.

api/src/main/kotlin/io/spine/protodata/settings/SettingsDirectory.kt
alexander-yevsyukov: As discussed vocally, we keep this API for the time being, hoping that adoption in Validation and McJava would give us more feedback on how we can improve it.

PR #206, opened 2024-01-24
Introduce `CreateSettingsDirectory` task

alexander-yevsyukov:
This PR introduces `CreateSettingsDirectory` task type which simply ensures that a directory for ProtoData settings files exists.

The task is needed for these reasons:
 * We need some code for ensuring the directory _when_ there are no other tasks that create setting files.
 * Having a task for this is the way recommended by Gradle and implied by the rules of validating `@InputDirectory` properties.
 * Should the user need to configure the settings directory, the task property is the place.

### Other notable changes
 * The property `settingsDir` was removed from `CodegenSettings` interface and `Extension` class accordingly.
 * Constants for directory names were consolidated under the `Directories` object.
 * Documentation tasks are disabled for `test-env` module.
 * `LaunchProtoData.requestFile` property became `val` and got type `RegularFileProperty` (instead of `Provider<RegularFile>`) for simplicity.
 * `LaunchProtoData.settingsDir` also became `val` an got type `DirectoryProperty` for the same reason.
 * `LaunchProtoData` task now depends on `CreateSettingsDirectory`.
 * The property `LaunchProtoData.optionProviders`  was removed as it's not used for quite some time.

### DIFF
--- a/buildSrc/src/main/kotlin/DokkaExts.kt
+++ b/buildSrc/src/main/kotlin/DokkaExts.kt
@@ -195,3 +195,20 @@ fun DokkaTask.isInPublishingGraph(): Boolean =
             startsWith("publish") && !startsWith("publishToMavenLocal")
         }
     }
+
+/**
+ * Disables Dokka and Javadoc tasks in this `Project`.
+ *
+ * This function could be useful to improve build speed when building subprojects containing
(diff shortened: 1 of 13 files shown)


armiol (approved):
@alexander-yevsyukov LGTM with one suggestion.

PR #207, opened 2024-01-31
Add `file` parameter to `TypeAnnotation`  

alexander-yevsyukov:
This PR teaches `TypeAnnotation` annotate top-level class of a given Java file, as an alternative to passing class name.

Other notable changes:

### Added extension function
  * `SourceFile.isJava`
  * `SourceFileSet.hasJavaOutput` and `hasGrpcOutput`
  * `Class<*>.isJavaLang`

These extensions would be useful in code generation in McJava.

### Deprecations
 * `SourceFile.lineSplitter` was deprecated since it does the same as `CharSequence.lines()`.

### DIFF
--- a/codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/annotation/TypeAnnotationSpec.kt
+++ b/codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/annotation/TypeAnnotationSpec.kt
@@ -51,7 +51,7 @@ internal class TypeAnnotationSpec {
     }
 
     @Test
-    fun `reject both 'subject' and 'file' arguments`() {
+    fun `reject simultaneously 'subject' and 'file' arguments`() {
         assertThrows<IllegalArgumentException> {
             StubAnnotation(SuppressWarnings::class.java,
                 ClassName("foo", "bar"),
--- a/codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/annotation/TypeAnnotationSpec.kt
+++ b/codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/annotation/TypeAnnotationSpec.kt
@@ -51,6 +51,20 @@ internal class TypeAnnotationSpec {
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/annotation/TypeAnnotationSpec.kt
```diff
@@ -50,6 +50,20 @@ internal class TypeAnnotationSpec {
...
+        assertThrows<IllegalArgumentException> {
+            StubAnnotation(SuppressWarnings::class.java,
+                ClassName("foo", "bar"),
+                SourceFile.fromCode(Paths.get("stub", "Source.java"), """
+                    class Source { }    
+                    """.trimIndent()
+                )
+            )
+        }
+
```
armiol: This line is redundant.

codegen/java/src/test/kotlin/io/spine/protodata/codegen/java/annotation/TypeAnnotationSpec.kt
```diff
@@ -50,6 +50,19 @@ internal class TypeAnnotationSpec {
         }
     }
 
+    @Test
+    fun `reject both 'subject' and 'file' arguments`() {
```
armiol, line 54: Let's say "simultaneously set" instead of "both".

armiol (approved):
@alexander-yevsyukov LGTM with some comments to address. 

I also assume, the "mcClient" side of this ProtoData server will still ask end-users just for class names (by pattern), not for the file names as an alternative?

In any case, it's a matter of future changes, not of this PR.

alexander-yevsyukov:
> I also assume, the "mcClient" side of this ProtoData server will still ask end-users just for class names (by pattern), not for the file names as an alternative?

The `ClassPatternAnnotator` renderer in McJava would traverse files, converting their names to fully-qualified names, then matching to the pattern, and then request annotation of the top-level class, if matching.

The primary purpose of the new `file` parameter is to avoid unnecessary search by a class name.

PR #208, opened 2024-02-01
Simplify annotation check flow

alexander-yevsyukov:
This PR updates the `TypeAnnotation` class to simplify the check whether an annotation should be put into the code.

Previously, `shouldAnnotate()` method was called only when annotating multiple files. In some cases, the check is also needed when handling annotation of only one file (when calling `annotate(SourceFile)`). In such a case we need to see if a source code was _already_ annotated via another `Renderer` which delegates to `TypeAnnotation`. So, now `annotate()` calls `shouldAnnotate()` first.

### Other notable changes
  * A cache was introduced for avoiding repeated parsing when obtaining `PsiJavaFile` instance for a `Text` instance.
  * Kotlin-based line splitting is now used in `SourceAtLine` instead of using `TextFactory.lineSplitter()`. It is expected that `lineSplitter()` would be [deprecated ](https://github.com/SpineEventEngine/text/issues/7) in Text library soon.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
+++ b/api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
@@ -269,6 +269,9 @@ private constructor(
         return text(code)
     }
 
+    /**
+     * Adds an action to be executed before obtaining the [text] of this source file.
+     */
     internal fun whenRead(action: (SourceFile) -> Unit) {
         preReadActions.add(action)
         if (alreadyRead) {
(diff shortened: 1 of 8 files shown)


api/src/main/kotlin/io/spine/protodata/renderer/SourceFile.kt
```diff
@@ -269,6 +269,9 @@ private constructor(
         return text(code)
     }
 
+    /**
+     * Adds an action to be executed before obtaining the [text] of this source file.
+     */
     internal fun whenRead(action: (SourceFile) -> Unit) {
```
armiol, line 275: I understand "when read", as "when it has been read". So, after the operation is done. 

Can we rename it to `beforeRead`?

armiol (approved):
@alexander-yevsyukov LGTM, but please see my suggestion before merging.

PR #218, opened 2024-03-26
Publish sources and Javadoc with `gradle-plugin`

alexander-yevsyukov:
This PR adds instruction to publish source code .jar and Javadoc .jar when publishing ProtoData Gradle plugin.

### DIFF
--- a/gradle-plugin/build.gradle.kts
+++ b/gradle-plugin/build.gradle.kts
@@ -94,6 +94,11 @@ val functionalTest: Task by tasks.getting {
     }
 }
 
+java {
+    withSourcesJar()
+    withJavadocJar()
+}
+
 val pluginName = "protoDataPlugin"
 
 gradlePlugin {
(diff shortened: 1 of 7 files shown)


gradle-plugin/build.gradle.kts
```diff
@@ -94,6 +94,11 @@ val functionalTest: Task by tasks.getting {
     }
 }
 
+java {
+    withSourcesJar()
+    withJavadocJar()
```
armiol, line 99: We'll need to inspect that JAR, once it gets published. 

Especially, in the light of what we've discussed with @yevhenii-nadtochii recently.

armiol (approved):
LGTM.

gradle-plugin/build.gradle.kts
alexander-yevsyukov: I've checked. It contains Kotlin source code files — because there's no Java code in this module. Javadoc JAR looks like regular Javadoc. Nothing fancy. 

I did not add publishing Dokka JAR for the follow reasons:
 1. Make it happen faster because I need source code of the ProtoData Gradle plugin for my current debug tasks.
 2. As we've learned that Kotlin source code is essential for IDEA to show documentation.
 3. I guess we'll need a generic solution for Dokka docs produced for plugins anyway.

PR #221, opened 2024-04-01
Bring extensions from McJava

alexander-yevsyukov:
This PR brings extensions previously adopted in McJava when developing new code generation for `EntityState` messages.

Instead of having all the extensions in `AstExts.kt` and `Ast2Java.kt` files, which grew too big already, new convention was adopted:
 * Extensions are gathered in Kotlin files named after the pattern `<TypeName>Exts.kt`. For example, `EnumTypeExts.kt`.
 * Kotlin files get `JvmName` after the pattern `<TypeName>s`.
 * Extensions for standard types are named after the package they are in: `JavaLangExts.kt`. 
 * A `JvmName` for standard types is formed after the pattern `<PackageName>Types`. For example, `JavaLangTypes`.

### Other notable changes
 * **Extensions from `DescriptorToAstExts.kt` were moved to `DescriptorExts.kt`.** This file now gathers extensions for all Protobuf descriptor types, which differs from the approach we selected for ProtoData types. We do this for simplicity because not many new extensions are expected for the descriptors types.
 
 * **Tests from `AstExtsSpec` were separated by files** after the pattern `<Type>ExtsSpec` even if they cover extensions declared in `DescriptorExts.kt` file.

* **`Class.codeReference` extension was deprecated** in favor of `Class.reference`.


armiol (approved):
LGTM.

PR #222, opened 2024-04-02
More extensions from McJava

alexander-yevsyukov:
This PR brings more extensions previously adopted in McJava.


armiol (approved):
@alexander-yevsyukov LGTM.

PR #224, opened 2024-04-11
Fix deep scan of `MessageTypeDependencies`

alexander-yevsyukov:
This PR fixes the `MessageTypeDependencies` for the case of scanning message types used by those used in the fields.

Also the `MessageTypeDependencies.set` property was deprecated into favor of the `scan()` method which produces a copy of the scanned `Set`. This way we avoid holding a reference to the "guts" of an instance of `MessageTypeDependencies` which is not of any use after we obtained the dependencies.

### DIFF
--- a/api/src/testFixtures/proto/given/type/project.proto
+++ b/api/src/testFixtures/proto/given/type/project.proto
@@ -98,7 +98,7 @@ message NoColumnsView {
 message IdOnlyView {
     option (entity).kind = PROJECTION;
 
-    UserId id = 3;
+    UserId id = 1;
 }
 
 message UserId {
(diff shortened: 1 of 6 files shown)


api/src/testFixtures/proto/given/type/project.proto
```diff
@@ -0,0 +1,110 @@
...
+    option (entity).kind = PROJECTION;
+
+    UserId id = 1;
+    UserName name = 2;
+}
+
+message IdOnlyView {
+    option (entity).kind = PROJECTION;
+
+    UserId id = 3;
```
armiol, line 101: Please make it numbered "1".

armiol (approved):
LGTM with a comment to address.

PR #225, opened 2024-04-12
Add `MessageType.javaClassName(TypeSystem)` extension function

alexander-yevsyukov:
This PR adds `MessageType.javaClassName(TypeSystem)` extension function which is a shortcut for obtaining a Java class name for a message type. This extension supplements existing `MessageType.javaClassName(ProtoFileHeader)` extension allowing to avoid funding the header first.

Also, the dependency on `tool-base` was bumped.

### DIFF
--- a/java/src/main/kotlin/io/spine/protodata/java/MessageTypeExts.kt
+++ b/java/src/main/kotlin/io/spine/protodata/java/MessageTypeExts.kt
@@ -30,6 +30,7 @@ package io.spine.protodata.java
 
 import io.spine.protodata.MessageType
 import io.spine.protodata.ProtoFileHeader
+import io.spine.protodata.type.TypeSystem
 import java.nio.file.Path
 
 /**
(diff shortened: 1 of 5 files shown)


java/src/main/kotlin/io/spine/protodata/java/MessageTypeExts.kt
```diff
@@ -48,3 +49,16 @@ public fun MessageType.javaFile(accordingTo: ProtoFileHeader): Path =
  */
 public fun MessageType.javaClassName(accordingTo: ProtoFileHeader): ClassName =
     name.javaClassName(accordingTo)
+
+/**
+ * Obtains a clas name for the Java code generated for this message type.
```
armiol, line 54: "class"

armiol (approved):
@alexander-yevsyukov LGTM, but please address the typo.

PR #226, opened 2024-04-16
Add `FilePattern` type

alexander-yevsyukov:
This PR adds `FilePattern` type copying its definition from the Validation subproject. The goal is "lift" the type, making it more reusable, avoiding the dependency on Validation.

The nature of the `FilePattern` does not have much to do with the Validation. This type is also used in McJava code generation settings related to matching message types to file patterns. The current arrangement feels awkward.

### `FilePatternFactory` was introduced
The factory is meant to replace `io.spine.tools.mc.java.gradle.settings.PatternFactory` and `io.spine.tools.mc.java.gradle.settings.FilePatterns`.  The factory object approach was selected to allow shorter function names, avoiding `file` prefix, which we used to have in `FilePatterns` in McJava.

In Kotlin, we can easily use the imported functions avoiding the factory object name, or using the object name, if we have name clash with factory methods for other types with the same names.

In Java, we'd call the factory methods via `FilePatternFactory.INSTANCE` provided by new implementation of `io.spine.tools.mc.java.gradle.settings.MessageCodegenOptions.by()`.

### Other notable changes
 * `FilePattern.matches(MessageType)` extension function so that Validation and McJava can reuse it.
 * `File.name` ext. `val` was introduced to simplify obtaining just a file name (which includes the extension). It was possible to obtain a file name via `toPath().name.toString()`, but it was mouthful.

### DIFF
--- a/api/src/main/proto/spine/protodata/file.proto
+++ b/api/src/main/proto/spine/protodata/file.proto
@@ -65,4 +65,3 @@ message FilePattern {
         string regex = 3;
     }
 }
-
--- a/api/src/main/kotlin/io/spine/protodata/FilePatternExts.kt
+++ b/api/src/main/kotlin/io/spine/protodata/FilePatternExts.kt
@@ -24,45 +24,23 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-syntax = "proto3";
+package io.spine.protodata
 
-package spine.protodata;
+import io.spine.protodata.FilePattern.KindCase.KIND_NOT_SET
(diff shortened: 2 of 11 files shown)


api/src/main/proto/spine/protodata/file.proto
```diff
@@ -48,3 +48,21 @@ message File {
...
+        string suffix = 1;
+
+        // The path must start with this string.
+        string prefix = 2;
+
+        // The path must match this regular expression.
+        string regex = 3;
+    }
+}
+
```
armiol, line 68: This line is redundant.

armiol (approved):
LGTM with a minor comment.

PR #230, opened 2024-04-22
Introduce `PipelineSetup` testing utility

alexander-yevsyukov:
This PR initiates `protodata-testlib` artifact with the `PipelineSetup` utility.

Code generation process for a ProtoData plugin requires an instance of `Pipeline`. 
Its constructor accepts `CodeGeneratorRequest` instance.  
There are simplest cases of code generation (e.g. `TypeAnnotation` and derived classes) that do not require much from the information coming from `protoc`.  For such cases we can pass `CodeGeneratorRequest.getDefaultInstance()`. More sophisticated codegen needs a legitimate `CodeGeneratorRequest` because `Pipeline` produces numerous events about the Protobuf code out of it. 

`PipelineSetup` class can be created in two ways. The first way, through the primary constructor, is for the rare cases of having the generated code and `CodeGeneratorRequest` instance already available some way. The second, recommended, way is through the factory method `byResources`, which assumes having ProtoTap Gradle Plugin added to the project. Please see the documentation for the `PipelineSetup` class for more details.

### Other notable changes
 * The `SettingsDirectory.directory` property was renamed to `path` for better readability at the use sites.
 * Extensions for `Format` enum items were sorted in the order from most to least expected.
 * Improved documentation of the `main()` of the ProtoData `protoc` plugin.


api/src/main/proto/spine/protodata/settings.proto
```diff
@@ -70,21 +70,31 @@ extend google.protobuf.EnumValueOptions {
...
+    //
+    // See: https://protobuf.dev/programming-guides/techniques/#suffixes
+    // See: https://buf.build/docs/reference/inputs#binpb
+    //
+    PROTO_BINARY = 1 [(extension) = "binpb", (extension) = "pb", (extension) = "bin"];
 
     // A Protobuf message encoded in Protobuf JSON.
+    //
+    // Use this item instead of [JSON] for Protobuf messages stored in
+    // JSON format so that correct parser is selected for the file.
```
armiol, line 96: "the correct parser"

buildSrc/src/main/kotlin/io/spine/internal/gradle/kotlin/KotlinConfig.kt
```diff
@@ -58,6 +58,7 @@ fun KotlinCompile.setFreeCompilerArgs() {
             "-Xskip-prerelease-check",
             "-Xjvm-default=all",
             "-Xinline-classes",
+            "-Xexpect-actual-classes",
```
armiol, line 61: Let's discuss this one. I couldn't find any significant details on this option.

protoc/build.gradle.kts
```diff
@@ -43,10 +42,10 @@ tasks.jar {
             else -> zipTree(it)
         }
     })
-
-    // We should provide a classifier or else Protobuf Gradle plugin will substitute it with
```
armiol, line 47: I would keep this comment to `archiveClassifier.set("exe")`.

testlib/src/main/kotlin/io/spine/protodata/testing/PipelineSetup.kt
```diff
@@ -0,0 +1,276 @@
...
+ * 
+ * Here are the steps for creating an instance of `PipelineSetup` in your tests.
+ *
+ * #### 1. Add [`java-test-fixtures`](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures) plugin
+ * The plugin will create the `testFixtures` source set in which we will put proto files.
+ *
+ * Alternatively, you can put proto files under the `test` source set, but with `testFixtures`
+ * it is a bit more neat.
+ *
+ * ### 2. Add [Protobuf Gradle Plugin](https://github.com/google/protobuf-gradle-plugin)
```
armiol, line 81: Please keep the header level consistent. Here and below it should be `####`.

testlib/src/main/kotlin/io/spine/protodata/testing/PipelineSetup.kt
```diff
@@ -0,0 +1,276 @@
...
+ * would pick them automatically.
+ *
+ * If you're going to use a custom source set, please pass it to
+ * the [plugin settings](https://github.com/SpineEventEngine/ProtoTap?tab=readme-ov-file#using-plugin-settings).
+ *
+ * ### 4. Add a test method with two `@TempDir` parameters
+ * This step applies if you're using JUnit. We will need two directories: one is for storing
+ * settings for the ProtoData plugins we're going to test, and another is for the output of
+ * the code generation process:
+ * ```kotlin
```
armiol, line 98: Please allow empty lines above `kotlin` blocks. It's here and below.

testlib/src/main/kotlin/io/spine/protodata/testing/PipelineSetup.kt
```diff
@@ -0,0 +1,276 @@
...
+            return inputRoot
+        }
+    }
+}
+
+@VisibleForTesting
+internal fun Language.protocOutputDir(): String {
+    return when(this) {
+        Java, Kotlin -> name.lowercase()
+        TypeScript -> "ts"
```
armiol, line 270: This is in fact configurable. Since TypeScript is not a built-in, but rather a number of custom plugins, each of them will be able to set its own folder name.

So I would mention this in the docs, as a convention we introduce.

testlib/src/main/kotlin/io/spine/protodata/testing/PipelineSetup.kt
```diff
@@ -0,0 +1,276 @@
...
+        // It's not likely we have proto files in the output of `protoc` or ProtoData anytime soon.
+        // But let's cover this case of the mete-codegen tools that produce Protobuf code.
```
armiol, line 272: "meta"

armiol (changes requested):
@alexander-yevsyukov LGTM with some comments.

testlib/src/main/kotlin/io/spine/protodata/testing/PipelineSetup.kt
alexander-yevsyukov: Described in a section in the class doc.

testlib/src/main/kotlin/io/spine/protodata/testing/PipelineSetup.kt
alexander-yevsyukov: I've moved them all one level up, making nesting consistent.

armiol (approved):

PR #233, opened 2024-05-17
Create language subdirectory in `PipelineSetup.byResources()`

alexander-yevsyukov:
This PR changes the way the factory method `PipelineSetup.byResources()` work in relation to a output directory passed as the parameter.

Previously, we assumed that the directory is the direct target for the output files. It turned out that some code, in particular the `SourceFileSet.hasJavaOutput` extension relies on the precise name of the subdirectory (`java`) which was used by McJava to tell if the source file set deals with Java files.

In this PR we want to make it more reliable without dropping the `SourceFileSet.hasJavaOutput`. Instead, we do the following:

 1.  Now `PipelineSetup.byResources()` assumes that the directory passed as `outputRoot` is the root directory for all the languages handled by Protobuf. Correspondingly,  the method creates a subdirectory for the generated source code of the given `language` using the `protocOutputDir()` extension function.
 2. `SourceFileSet.hasJavaFiles` extension value is introduced to help the renderers decide if they really need to deal the generated code.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -926,7 +926,7 @@
 
 The dependencies distributed under several licenses, are used according their commercial-use-friendly license.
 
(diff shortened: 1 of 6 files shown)


armiol (approved):
@alexander-yevsyukov as discussed, let's set the version to `0.23.0`. And then LGTM.

PR #234, opened 2024-05-18
Keep parsed `PsiFile` under `SourceFile`

alexander-yevsyukov:
This PR improves the way we work with parsing Java code into an instance of `PsiJavaFile` for further traversal and modifications.

## The problem
Before this PR we used to have a cache with the association of `Text` instances to `PsiJavaFile` instances that were created when parsing these texts. The problem with this approach is that it is error prone because of the following. When we get an instance of `PsiJavaFile` we can use it for both navigation and modification. Nothing prevents us from doing the latter. So, in order to keep the match `Text` to `PsiJavaFile` consistent, we need to re-parse the same text so that we can modify the code. This diminish the usefulness of the cache and puts additional burden on its users.

## The solution
Now the `SourceFile` is responsible for holding the reference to `PsiFile`. 

Unlike before this PR (via `SourceFile.toPsi()` ext. `fun`), the reference does no not have the `PsiJavaFile` type because `SourceFile` could be in any language. The language — and correspondingly `FileType` —  is detected when `psi()` method is called for the first time. The private `fileType` property is responsible for keeping it for repeated parsing calls.

When the code is updated by the `overwrite()` call, the `PsiFile` instance is discarded.

## Changes to other API
 * API calls that used to accept `Text` instances now accept `String`. `InsertionPoint` interface methods were deprecated accordingly. It turns out that having yet another intermediate abstraction is not convenient. `SourceFile` is good enough for our needs, especially assuming that we're going to use PSI more.

 * The `SourceFile.code()` method — which was deprecated — is now back in favour, and the `text()` method is deprecated.

 * `SourceFileSet` got the `project` property, which is lazily evaluated from `Environment.project`. This is a little step towards making our API more project-oriented, with the goal of being able to analyse the whole file tree of a software project based on Spine, eventually.

## Other notable changes
 * The `api` module now depends on `Spine.psiJavaBundle` at `api` level. Since we expose `PsiFile` — and indirectly — `PsiJavaFile` and all the “jungle” surrounds these types, it is more ergonomic for the users to get it all when adding ProtoData API module to their project.
 
## Some thoughts about future work
The whole story about the relation of `InsertionPoint` and `SourceFile` needs more attention. Now the points "hang" in the air. They insert into coordinates in a `String`. PSI looks like a way to make it more reasonable with coordinates being not "x,y" in a test, but `PsiElement` in a `PsiFile`. Now we have to deal with `Text` instances and its extensions just to be able to get the insertion coordinates.

Works in McJava, which drive the recent modifications in ProtoData, should give us more feedback on where improvements of the `InsertionPoint` API and other parts should be directed.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -39,6 +39,7 @@ dependencies {
         Spine.text,
         Spine.CoreJava.server,
         Spine.toolBase,
+        Spine.psiJavaBundle
     ).forEach {
         api(it)
     }
--- a/api/src/main/kotlin/io/spine/protodata/renderer/CoordinatesFactory.kt
+++ b/api/src/main/kotlin/io/spine/protodata/renderer/CoordinatesFactory.kt
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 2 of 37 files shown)


alexander-yevsyukov:
@armiol, this PR depends on [#87 in `tool-base`](https://github.com/SpineEventEngine/tool-base/pull/87). I've used it locally when working on this one. Just FYI.

armiol (approved):
@alexander-yevsyukov LGTM with a minor comment.

Also, regarding the `(x,y)`-insertion points. They cover all cases, for which PSI does not exist. So right now we need them for any language except Java. And even in the future, C# and other similar languages will probably not have PSI support.

PR #241, opened 2024-05-31
Store code generation request in JSON also

alexander-yevsyukov:
This PR teaches ProtoData to store `CodeGeneratorRequest` in JSON format in addition to the binary file. JSON output is needed for diagnostics and debugging purposes.


protoc/src/main/kotlin/io/spine/protodata/protoc/Plugin.kt
```diff
@@ -56,6 +59,10 @@ public fun main() {
 
     requestFile.writeBytes(request.toByteArray(), CREATE, TRUNCATE_EXISTING)
 
+    val requestFileInJson = requestFile.run { resolveSibling("$nameWithoutExtension.pb.json") }
```
armiol, line 62: I thought we wanted this thing to be simplified with those `Path` extension.

armiol (approved):
@alexander-yevsyukov LGTM, but please see my comment.

PR #243, opened 2024-06-01
Migrate `protoc` plugin to `CodeGeneratorRequestWriter`

alexander-yevsyukov:
This PR migrates the implementation of ProtoData `protoc` plugin to `CodeGeneratorRequestWriter` recently introduced in `base`.

`config` was also updated.


buildSrc/src/main/kotlin/jacoco-kotlin-jvm.gradle.kts
```diff
@@ -25,6 +25,11 @@
  */
 
 import io.spine.internal.gradle.buildDirectory
+import java.io.File
```
armiol, line 28: The changes in this file are a mystery to me. How all of that worked previously without imports, and why the warning suppression is no longer needed, I have no idea.

armiol (approved):

PR #258, opened 2024-09-10
Bring back `IncrementGuard`

armiol:
This changeset updates the build scripts, so that we start checking for the possible duplication in the versions of the artifacts we are going to publish.

Previously, for some reason, `ProtoData` modules, for the most part, did not check for that.

The Maven repository used is the GCR (`https://europe-maven.pkg.dev/spine-event-engine`), which as checked manually, has all the versions of recent ProtoData artifacts published, and hence, is eligible for being a target for the version checks.

The ProtoData version is set to `0.60.1`.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #260, opened 2024-09-13
Implement `ClassName.outer()` function

Oleg-Melnik:
This PR adds `ClassName.outer()` function, which allows obtaining a new `ClassName` with the last element removed from the list of simple names.

The ProtoData version is set to `0.60.3`.

### DIFF
--- a/java/src/main/kotlin/io/spine/protodata/java/ClassName.kt
+++ b/java/src/main/kotlin/io/spine/protodata/java/ClassName.kt
@@ -107,7 +107,7 @@ public class ClassName(
      * The method is useful for obtaining names for outer classes,
      * e.g. `Message` from `Message.Builder`.
      *
-     * @throws IllegalStateException when the current class name is the top level class name.
+     * @throws IllegalStateException When the current class name is the top level class name.
      */
     public fun outer(): ClassName {
         check(simpleNames.size > 1) {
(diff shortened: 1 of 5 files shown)


java/src/main/kotlin/io/spine/protodata/java/ClassName.kt
```diff
@@ -101,6 +101,21 @@ public class ClassName(
     public fun nested(simpleClassName: String): ClassName =
         ClassName(packageName, simpleNames + simpleClassName)
 
+    /**
+     * Obtains a new `ClassName` with the last element removed from the list of simple names.
+     *
+     * The method is useful for obtaining names for outer classes,
+     * e.g. `Message` from `Message.Builder`.
+     *
+     * @throws IllegalStateException when the current class name is the top level class name.
```
alexander-yevsyukov, line 110: Please capitalise `when`. Have a look at the [documentation conventions](https://github.com/SpineEventEngine/documentation/wiki/KDoc-and-Dokka#documenting-a-type) we use.

java/src/main/kotlin/io/spine/protodata/java/ClassName.kt
```diff
@@ -101,6 +101,21 @@ public class ClassName(
...
+     */
+    public fun outer(): ClassName {
```
alexander-yevsyukov, line 112: If would be easier to deal with this API if we return `ClassName?` instead of throwing the exception.

For example, with a nullable output it would be possible to climb up the nested classes. With the current approach you'd have to catch exceptions to do that.

alexander-yevsyukov (changes requested):
Please see my comments.

alexander-yevsyukov (approved):
LGTM

PR #265, opened 2024-09-19
Repackaging and render types from McJava and 

alexander-yevsyukov:
This PR changes the package structure of the `api` and `java` modules. Also it adds rendering types recently adopted in McJava to the new `io.spine.protodata.java.render` package.

## Changes to the package structure
 * The package `io.spine.protodata.renderer` was renamed to `render` to better serve the content of the package which contains not only `Renderer` classes.
 * The types related to parsing of Protobuf source code (e.g. coming from `ast.proto` and the like) and corresponding Kotlin extensions were moved under the `io.spine.protodata.ast` package.
 * Extensions that transform Protobuf `Descriptor` instances into ProtoData AST types were moved under the `io.spine.protodata.protobuf` package.
 * The Java code produced out of `value.proto` and associated extensions now below to `io.spine.protodata.value` package.
 * `CodegenContext`, `ContextAware`, and `Member` were moved under the `io.spine.protodata.context` package.
 * `ConfigurationError` exception class was moved under `io.spine.protodata.plugin`. Only plugin-related classes now throw it. Other errors were migrated to `error()` calls.
 * Proto declaration `FilePattern` moved into a dedicated file to make the usage explicit in imports.
 * The proto type `InsertionPointPrinted` moved into the file `render_events.proto` with the target Java package `io.spine.protodata.render.event`.
 * The proto type `Actions` moved into `render_actions.proto` file with the target Java package `io.spine.protodata.render`. The file was named with the `render_` prefix  to make it come later in alphabetical sorting. We want this file listed after `ast.proto` to simplify learning the proto types.
 * The target Java package for `style.proto` is now `io.spine.protodata.style`.
 * The `JavaRenderer` class moved into the `io.spine.protodata.java.render` package.

## Types from McJava
The following general purpose rendering types recently created and modified in McJava were added to the `io.spine.protodata.java.render`:
 * `AbstractRenderer` 
 * `BaseRenderer`
 * `CreateInterface`
 * `CreateNestedClass`
 * `DirectMessageAction`
 * `ImplementInterface`
 * `MessageAction`
 * `RenderActions`
 * `TuneNestedClass`
 * `TypeActions`
 * `TypeListActions`
 * `TypeListRenderer`
 * `TypeRenderer`
 * `SuperInterface` proto type and corresponding Java code.

## Deprecations
 * The class `RejectionThrowableConvention` was deprecated encouraging its move to McJava.


api/src/main/kotlin/io/spine/protodata/settings/Formats.kt
```diff
@@ -54,11 +53,11 @@ public val Format.extensions: List<String>
 /**
  * Obtains a [Format] from the file extension of the given configuration file.
  *
- * @throws ConfigurationError if the format is not recognized
+ * @throws IllegalStateException if the format is not recognized
```
armiol, line 56: I am not sure how we agreed on dealing with `@throws` and capitalization. If we went for the simple rule, then it should be "If the format..." (and similar to the capitalized version you have in `Parser.kt`).

If we agreed on not capitalizing `@throws`, then `Parser.kt` has this issue, because in it, it is "If..."

java/src/main/kotlin/io/spine/protodata/java/render/TypeListRenderer.kt
```diff
@@ -0,0 +1,65 @@
...
+import io.spine.protodata.render.SourceFile
+import io.spine.tools.code.Java
+
+/**
+ * An abstract base for Java renders handling message types.
+ *
+ * This class applies multiple render actions to multiple types.
+ * For applying rendering actions to one type, please see [TypeRenderer].
+ *
+ * @param V the type of the view state which gathers messages types served by this renderer.
```
armiol, line 42: Here and below it should be "The".

api/src/main/kotlin/io/spine/protodata/render/SourceFile.kt
```diff
@@ -24,16 +24,16 @@
...
 import com.google.common.annotations.VisibleForTesting
 import com.intellij.openapi.fileTypes.FileType
 import com.intellij.openapi.fileTypes.FileTypeRegistry
 import com.intellij.psi.PsiFile
 import com.intellij.psi.PsiFileFactory
 import com.sksamuel.aedile.core.cacheBuilder
-import io.spine.protodata.InsertedPoints
-import io.spine.protodata.file
+import io.spine.protodata.ast.file
+import io.spine.protodata.render.SourceFile.Companion.fromCode
```
armiol, line 36: This is suspicious, since it's a new import from this very file, and nothing else is changed.

I have deleted this import in my IDEA, and nothing bad happened. I think, it's just a consequence of excessive "help", provided by IDE. And it is here, because it is used in the documentation.

Please review.

armiol (approved):
@alexander-yevsyukov LGTM with a few comments.

PR #270, opened 2024-10-09
Allow more than one cardinality when calculating message type dependencies

alexander-yevsyukov:
This PR extends the `MessageTypeDependencies` class allowing to pass more than one `CardinalityCase` for filtering. 

The change is needed to allow gathering combination of `SINGLE` and `ONEOF_NAME` cardinalities. It is needed to gather deeper dependencies when nested fields has an alternative cardinality than a field of with the outer message type.

Convenience `MessageTypeDependencies` constructors covering non-combination cases were also provided.

## Other notable changes
 * Previously deprecated `set` property and `scan()` method of `MessageTypeDependencies` were removed.
 * Added `EventContextDependenciesTest` which documents expected dependencies of the `spine.core.EventContext` type. 
 * `buildSrc` was updated from latest `config`.


alexander-yevsyukov:
@yevhenii-nadtochii, FYI. These changes will be shortly applied to McJava to fix the issue with `oneof` handling for `Field` classes.

armiol (approved):

PR #272, opened 2024-10-10
Support field reference in `Value`

alexander-yevsyukov:
This PR extends `Value` type with the ability to carry a reference to a value in another field. 

One of the usage scenarios for such a feature is validation options using references to another field. For example, consider the following proto type `Range` in which the `min_value` can have a value not bigger than `max_value`.
```proto
message Range {
    int32 min_value = 1 [(max).value = "max_value"];
    int32 max_value = 2;
}
```
A reference could made to a nested field using dot-separated fields names in the path.

## Changes to `Value` and related code
 * The type `Reference` was introduced.
 * The type `Value` got the new field `reference`.
 * `MinOption` and `MaxOption` got extensions `parse()` for obtaining corresponding instances of `Value`.

## Other notable changes
 * The event `FieldOptionDiscovered` got a new field called `subject` with type `Field`. The new field is aimed to replace now deprecated fields `type` and `field`.
 * The creation of a singular value via `FieldDescriptor` made stricter, failing with `IllegalStateException` when unsupported `JavaType` enum item is encountered. This is an unlikely case, but it's still better to throw rather than silently return `NULL`.
 * `ProtobufSourceFile` type got extension for finding a message type by its simple name and possible enclosing type names.
* `TypeSystem` got an extension `resolve()` which allows to obtain a `Field` instance for a field of a message type specified by `FieldPath`.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/value/Options.kt
+++ b/api/src/main/kotlin/io/spine/protodata/value/Options.kt
@@ -30,9 +30,11 @@ package io.spine.protodata.value
 
 import io.spine.base.FieldPath
 import io.spine.base.FieldPathConstants
+import io.spine.base.joined
 import io.spine.option.MaxOption
 import io.spine.option.MinOption
 import io.spine.protodata.ast.Field
+import io.spine.protodata.ast.name
 import io.spine.protodata.ast.qualifiedName
 import io.spine.protodata.type.TypeSystem
 import io.spine.protodata.type.resolve
@@ -48,7 +50,7 @@ public fun MinOption.parse(field: Field, typeSystem: TypeSystem): Value =
     OptionValue("min", value, field, typeSystem).parse()
 
 /**
- * Parses the value of this `(min)` option into an instance of [Value].
+ * Parses the value of this `(max)` option into an instance of [Value].
  *
  * The value could be an integer, double, or a reference to another field.
  */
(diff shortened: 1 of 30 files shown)


api/src/main/kotlin/io/spine/protodata/value/Options.kt
```diff
@@ -0,0 +1,106 @@
...
+                    type = field.type
+                    target = ensureField(FieldPath(value))
+                }
+            }
+            else -> error("${optionPath()} has the value with unexpected format (`$value`).")
+        }
+    }
+
+    private fun optionPath() =
+        "The `($optionName)` option declared in the field `${field.name.value}` of" +
```
armiol, line 91: Maybe, "for the field"?

armiol (approved):
@alexander-yevsyukov LGTM with a minor comment to consider.

api/src/main/kotlin/io/spine/protodata/value/Options.kt
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Parses the value of this `(min)` option into an instance of [Value].
+ *
+ * The value could be an integer, double, or a reference to another field.
+ */
+public fun MinOption.parse(field: Field, typeSystem: TypeSystem): Value =
+    OptionValue("min", value, field, typeSystem).parse()
+
+/**
+ * Parses the value of this `(min)` option into an instance of [Value].
```
yevhenii-nadtochii, line 51:
```suggestion
 * Parses the value of this `(max)` option into an instance of [Value].
```

api/src/main/proto/spine/protodata/value.proto
```diff
@@ -166,3 +171,19 @@ message MapValue {
...
+
+// A reference to a value in the field specified in the [target] property.
+//
+// Typically such a reference is obtained by parsing an option of a "source" field.
+//
+message Reference {
+
+    // The type of the referenced value.
+    //
+    // If the reference is obtained by parsing a "source" field option, it is the type of the field.
```
yevhenii-nadtochii, line 183: I don't fully get this sentence.

Did you mean "If ..., it is the type of the **referenced** field."? But then what if not?

java/src/test/kotlin/io/spine/protodata/java/MessageTypeExtsSpec.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.spine.protodata.protobuf.toPbSourceFile
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+@DisplayName("`MessageType` extensions for Java should")
+internal class MessageTypeExtsSpec {
+
+    @Nested inner class
+    `find its class` {
```
yevhenii-nadtochii, line 47: We don't need a nested class here. The second test case is negative.

yevhenii-nadtochii (approved):
LGTM with a couple suggestions.

armiol (approved):

PR #274, opened 2024-10-15
Override `context` and `typeSystem` properties for `RenderAction`

yevhenii-nadtochii:
This PR makes `context` and `typeSystem` properties non-nullable for `RenderAction` and its inheritors. The action registers within the context during the initialization, making the mentioned properties available.

Also, it moves `MessageType.javaClass()` extension, which emerged in `mc-java`.


yevhenii-nadtochii:
@alexander-yevsyukov PTAL

alexander-yevsyukov (approved):
LGTM

PR #278, opened 2024-10-18
Turn `Plugin` into the abstract class

alexander-yevsyukov:
This PR introduced a breaking API change turning `Plugin` which used to be an interface into the abstract class.

This change is needed to make the `Plugin` API more predictable to the usage sites in terms of consistency of returned lists and sets. For example, before this PR some of the plugins could return a new list of `Renderer`s each type the method `renderers()` was called. while other plugins returned the same list each time.

The benefits of having `Plugin` as the abstract class:
 1. It's easier to grasp for new users.
 2. Predictable behaviour for arranging traversal of parts (e.g. injecting a dependency to all the `Renderer`s).
 3. Predictable behaviour for arranging mutual work of the plugins, e.g. for potential sorting of `Renderer`s by their dependencies on each other. 

## Changes to `Plugin` API
 * The `Plugin` interface turned into abstract class, exposing public properties `renderers`, `views`, `viewRepositories`, and `policies`. Public methods for accessing those things were removed.
 * Correspondingly, the `AbstractPlugin` class was removed.
 * Stub types that mixed `Renderer` and `Plugin` interfaces were transformed into pairs of classes.

## Other notable changes
 * The class `io.spine.protodata.backend.ImplicitPluginWithRenderers` was renamed to `RenderingTestbed` and moved under the `testlib` module (under the package `io.spine.protodata.testing`).


armiol (approved):
@alexander-yevsyukov LGTM.

PR #283, opened 2024-10-29
Introduce `FieldType`, improve `Field` and `Reference` API

alexander-yevsyukov:
This PR introduces improvements to the field information, introducing new API.

New types are described below.

## `FieldType`
This is a new type which provides sufficient information about a type of a field. Previously, we used the combination of  the `Type` message and the `Field.cardinality` `oneof` which was 1) awkward and 2) could not be used outside of the `Field` type.
 * Extension functions converting `FieldDescriptor` to `FieldType` were added.
 * Extension functions and values were added.
 * A new `MapEntryType` was introduced to be the type of the `map` case under `FieldType.kind` `oneof`.

## `Cardinality`
This is a new `enum` which is replacing `oneof` `Field.cardinality`. The problem with `Field.cardinality` — apart from the those associated with the needs for introducing `FieldType` — is that it was used for telling if a field is a part of `oneof`. Such a "mixture" of meanings proved to be error prone.

## Changes to the `Field`
 * The `candinality` `oneof` was removed.
 * The `type` field now is of `FieldType`.
 * A new field called `enclosing_oneof` was introduced to replace the `oneof_name` field which was under the `cardinality` `oneof`.
 * The extension `isRepeated` was deprecated because it is misleading. Its name exposes implementation details of `map`s confusing it with the `repeated` declaration for fields.
 * The extension function `toType()` was introduced to help obtaining a field type as `Type` instance, if such a conversion is applicable.

## Other notable changes
 * The API of `ValueConverter` was improved to straighten conversion of the data enclosed into a `Value`.
 * `ValueConverter` now can transform `Reference` into Java code.
 * `MessageTypeDependencies` was adjusted to work with new the `Cardinality` enum.
 * `InstanceScope` expression object was introduced to allow calling message methods without referring to `this`.


alexander-yevsyukov:
@armiol, just FYI, I moved to newer API, and the build passes. 

I'm not quite happy with the longer field names (e.g., `field_type` instead of `type`) I had to "invent" in order to arrange the migration.

I'd like to discuss the next step as to removal of the existing deprecated API sometime tomorrow.

armiol (approved):

PR #284, opened 2024-11-01
Introduce `TypeBase` interface

alexander-yevsyukov:
This PR introduces the `TypeBase` interface, which allows to have common properties and extensions when both `Type` and `FieldType` are primitive, message, or enum. This allows to avoid unnecessary conversion from `FieldType` to `Type`.

## In details
 * The `TypeBase` interface is a `@GeneratedMixin` which covers the most interesting API common for `Type` and `FieldType`.
 * Extensions migrated to use `TypeBase` as the receiver when it's applicable.
 * `TypeNameOrBuilder.isAny` extension was added to be used in `TypeBase.isAny`.

## `FieldType` extensions
 * `isSingular` extension was introduced to simplify checks for being not a map, nor a list.
 * `extractMessageType()` — obtains a `MessageType` which this field type refers to directly or via the type of list or map values.
 * `extractPrimitiveType()` — similar to `extractMessageType()` but for `PrimitiveType`.

## Other notable changes
 * Test suites for `Type` and `FieldType` extensions were renamed to avoid `Exts` infix to cover the existence of the common base.
 * `package-info.java` file was added for the `io.spine.protodata.ast` to avoid the warning in the generated code.

### DIFF
--- a/api/src/main/java/io/spine/protodata/ast/package-info.java
+++ b/api/src/main/java/io/spine/protodata/ast/package-info.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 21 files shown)


api/src/main/java/io/spine/protodata/ast/package-info.java
```diff
@@ -0,0 +1,38 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * This package provides the data types and functions needed to work
+ * with the structure of Protobuf files.
+ */
+
```
armiol, line 31: I would remove this line.

api/src/main/kotlin/io/spine/protodata/ast/TypeBase.kt
```diff
@@ -0,0 +1,81 @@
...
+package io.spine.protodata.ast
+
+import com.google.protobuf.Message
+
+/**
+ * The common interface for [Type] and [FieldType].
+ */
+public interface TypeBase: Message {
+
+    public fun hasMessage(): Boolean
```
armiol, line 36: Let's write something on this group of methods. Although there are three of them, they look lonely.

api/src/main/kotlin/io/spine/protodata/ast/TypeBaseExts.kt
```diff
@@ -0,0 +1,81 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata.ast
+
+import io.spine.protodata.type.TypeSystem
+import io.spine.string.shortly
+
+/**
+ * Obtains a simple name of the type if represents a message or an enum.
```
armiol, line 33: I guess, "if it represents".

armiol (approved):
@alexander-yevsyukov LGTM with a few documentation-related comments.

armiol (approved):

PR #285, opened 2024-11-03
Add `FieldType` extensions adopted in Validation

alexander-yevsyukov:
This PR adds `FieldType` extensions there were recently [adopted in Validation](https://github.com/SpineEventEngine/validation/pull/141).

## Breaking changes
* `FieldTypeExts.kt` got `@file:JvmName("FieldTypes")` which requires updating imports in Java code.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/ast/FieldTypeExts.kt
+++ b/api/src/main/kotlin/io/spine/protodata/ast/FieldTypeExts.kt
@@ -24,6 +24,8 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+@file:JvmName("FieldTypes")
+
 package io.spine.protodata.ast
 
 import io.spine.protodata.ast.Cardinality.CARDINALITY_LIST
--- a/api/src/main/kotlin/io/spine/protodata/ast/FieldTypeExts.kt
+++ b/api/src/main/kotlin/io/spine/protodata/ast/FieldTypeExts.kt
@@ -49,24 +49,33 @@ public val FieldType.name: String
         ENUMERATION -> enumeration.qualifiedName
         PRIMITIVE -> primitive.protoName
         LIST -> "repeated ${list.name}"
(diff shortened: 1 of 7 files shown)


api/src/main/kotlin/io/spine/protodata/ast/FieldTypeExts.kt
```diff
@@ -161,3 +179,43 @@ public fun FieldType.extractPrimitiveType(): PrimitiveType? = when {
  */
 private fun Type.maybePrimitiveType(): PrimitiveType? =
     if (isPrimitive) primitive else null
+
+/**
+ * Extracts direct or element type information from this field type.
```
armiol, line 184: I would make it even easier to read:
```
Extracts the information about the direct type or the type of elements from this field type.
```

api/src/main/kotlin/io/spine/protodata/ast/FieldTypeExts.kt
```diff
@@ -161,3 +179,43 @@ public fun FieldType.extractPrimitiveType(): PrimitiveType? = when {
...
+ * @see extractTypeName
+ * @see extractMessageType
+ */
+public fun FieldType.extractType(): Type = when {
+    isMessage -> toType()
+    isEnum -> toType()
+    isPrimitive -> toType()
+    isMap -> map.valueType
+    isList -> list
+    else -> error("Cannot get type info from the field type ${this.shortly()}.")
```
armiol, line 195: Let's also format the type name:
```
"Cannot get type info from the field type `${this.shortly()}`."
```

version.gradle.kts
```diff
@@ -32,4 +32,4 @@
  *
  * For dependencies on Spine SDK module please see [io.spine.dependency.local.Spine].
  */
-val protoDataVersion: String by extra("0.65.1")
+val protoDataVersion: String by extra("0.65.2")
```
armiol, line 35: Given we introduced some breaking changes (as per the PR description), I would say it should be `0.66.0`.

armiol (approved):
@alexander-yevsyukov LGTM in general. Please see my suggestions.

PR #286, opened 2024-11-07
Extend Expression API to cover more codegen cases

yevhenii-nadtochii:
This PR extends Expression API to cover more use cases. 

Please take a look on [this](https://miro.com/app/board/uXjVLFfLAzU=/) Miro board for a bigger picture of the suggested changes.

### Generic argument for `Expression`

Now, instances of `Expression` have to declare the type of the result they yield. This helps to read and understand the code. Generified expressions also add more safety and allow building hierarchies of classes, which add up bits and bites to the generated code much safer.

Usage before:

```java
var ten = Expression("5 + 5");
```

Usage after:

```java
var ten = Expression<Int>("5 + 5");
```

Though, it is not possible to distinguish between a boxed and a non-boxed primitives using generics. And for the expression above, it is not clear where the expression returns an object or a primitive. As for now, we are ok with this.

`Expression.kt` file has been split in more files. It was starting to get too big to grasp.

### Extended hierarchy of `JavaTypeName`

This hierarchy now has classes to represent arrays, generified classes and type variables. `ClassOrEnumName` was merged to `ClassName` for simplicity. `EnumName` now is a child of `ClassName`.

There's no support for wildcards or type variable bounds. Though, `JavaTypeName` can be used for this.

### Statements and fields

`Statement` and `FieldDeclaration` are new sub-hierarchies for working with local variables and fields. The base `Statement` is instantiable.


java/src/main/kotlin/io/spine/protodata/java/Variable.kt
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Declares a local Java variable.
+ *
+ * The variable declaration is performed in Java 11 style, without
+ * an explicit variable type being specified.
+ *
+ * An example of the declared variable:
+ *
+ * ```
+ * val five = Variable<Int>("five", "5 + 5")
```
alexander-yevsyukov, line 38: Can we have it just `Var<Int>` and call the var `ten`?

java/src/main/kotlin/io/spine/protodata/java/Variable.kt
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Provides a read access to the variable with the given name.
+ *
+ * A variable of type [T] is also an [Expression] of the same type.
+ *
+ * @param T The type of the variable.
+ * @param name The name of the variable.
+ */
+public class VariableAccess<T>(name: String) : ArbitraryExpression<T>(name)
```
alexander-yevsyukov, line 63: `ReadVar<T>`?

java/src/main/kotlin/io/spine/protodata/java/MethodCall.kt
```diff
@@ -0,0 +1,124 @@
...
+ * @param T The method returned type.
+ *
+ * @param scope The scope of the method invocation: an instance receiving the method call, or
+ *   the name of the class declaring a static method.
+ * @param name The name of the method.
+ * @param arguments The list of the arguments passed to the method.
+ * @param generics The list of the type arguments passed to the method.
+ */
+public open class MethodCall<T> @JvmOverloads constructor(
+    private val scope: JavaElement,
```
yevhenii-nadtochii, line 44: I didn't change it. Though, it can be expanded to the notation of a receiver. A receiver can be either `Expression<R>` or `JavaElement` (some methods are invoked upon class names).

yevhenii-nadtochii:

This branch is in the failed state because I didn't push the commented out tests.

java/src/main/kotlin/io/spine/protodata/java/Expression.kt
```diff
@@ -28,364 +28,51 @@
...
-    if (isEmpty()) "" else "<${joinToString()}>"
-
-/**
- * Formats these expressions as method parameters, not including the brackets.
+ * @param T The type of the returned value.
+ * @param code The Java code denoting an expression.
  */
-private fun List<Expression>.formatParams() =
-    joinToString { it.toCode() }
+public open class ArbitraryExpression<T>(code: String): ArbitraryElement(code), Expression<T>
```
alexander-yevsyukov, line 78: Can we have `AnExpression` and `AnElement` just for brevity?

alexander-yevsyukov: Another question is, why do we need an open class which does not have anything apart from the value?

java/src/main/kotlin/io/spine/protodata/java/ExpressionTypes.kt
```diff
@@ -0,0 +1,37 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata.java
+
+/**
+ * An [Expression] return type stub that denotes a `null`.
+ */
+public object JavaNull
```
alexander-yevsyukov, line 32: Why does not it implements `Expression`?

java/src/main/kotlin/io/spine/protodata/java/MethodCall.kt
alexander-yevsyukov: We probably need a receiver which would allow avoiding `this.`, for example, when making a method call in the context where we can avoid having such a prefix.

java/src/main/kotlin/io/spine/protodata/java/MethodCall.kt
```diff
@@ -0,0 +1,124 @@
...
+    name: String,
+    arguments: List<Expression<*>> = listOf(),
+    generics: List<ClassName> = listOf()
+) : ArbitraryExpression<T>(
+    "${scope.toCode()}.${generics.genericTypes()}$name(${arguments.formatParams()})"
```
alexander-yevsyukov, line 49: This is the place where we can apply the notion of the receiver and referencing it via an empty string or `<receiver>.` before the method call.

java/src/main/kotlin/io/spine/protodata/java/Variable.kt
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Provides a read access to the variable with the given name.
+ *
+ * When being read, a variable of type [T] is also an [Expression]
+ * of the same type.
+ *
+ * @param T The type of the variable.
+ * @param name The name of the variable.
+ */
+public class ReadVariable<T>(name: String) : ArbitraryExpression<T>(name)
```
alexander-yevsyukov, line 71: In the current arrangement with `Variable` and `ReadVariable` we don't have a thing for the variable itself. The first is initialising the variable, and another is for reading it. But what is variable? 

Can a variable be a receiver for a method call? It probably should it if represents an object.

alexander-yevsyukov: I see this is getting complicated and we risk to start implementing Java grammar. I don't have a solution now. It's just an observation.

alexander-yevsyukov (commented):
Please see my comments.

armiol (changes requested):
@yevhenii-nadtochii as per our discussions yesterday, let's address the comments.

yevhenii-nadtochii:
@armiol @alexander-yevsyukov PTAL and let's discuss.

1. `Expression` and `Statement` are classes now.
2. `JavaElement` (future `JavaCode`) remained an interface because there are abstract classes between `JavaElement` and `ClassName`/`EnumName`.

java/src/main/kotlin/io/spine/protodata/java/KeywordExpressions.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * An expression that evaluates to `null`.
+ */
+public object Null : Expression<Null>("null")
+
+/**
+ * An expression that evaluates to nothing.
+ */
+public object Void : Expression<Void>("void")
```
alexander-yevsyukov, line 39: The questions we had with @armiol on this, Why don't we use `java.lang.Void` here?

java/src/main/kotlin/io/spine/protodata/java/VarInit.kt
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Declares a local Java variable.
+ *
+ * The variable declaration is performed in Java 11 style, without
+ * an explicit variable type being specified.
+ *
+ * An example of a variable declaration:
+ *
+ * ```
+ * val ten = Variable<Int>("ten", "5 + 5")
```
alexander-yevsyukov, line 38: It should be `VarInit<Int>` to match the new name.

java/src/main/kotlin/io/spine/protodata/java/VarInit.kt
```diff
@@ -0,0 +1,67 @@
...
+    public fun read(): Var<T> = Var(name)
+}
+
+/**
+ * Provides a read access to the variable with the given name.
+ *
+ * @param T The type of the variable.
+ * @param name The name of the variable.
+ */
+public class Var<T>(name: String) : Expression<T>(name)
```
alexander-yevsyukov, line 67: The remaining question here: "Do we need a class of the expression for just declaring a variable?"

If we do, then how do we name it?

alexander-yevsyukov: Another question related to this class (and its name) is this: "We have two classes that deal with variables via its name. Why don't they have a common base which encapsulates this concept?"

alexander-yevsyukov: I mean, it's probably `class Var<T>(public val value: String)`, then `class ReadVar<T>(name: String) : Var<T>`, and `VarDecl<T>(name: Strig) : Var<T>`.

alexander-yevsyukov (commented):
Please see my comments and questions. Let's discuss tomorrow with @armiol when time permits.

yevhenii-nadtochii:
@armiol @alexander-yevsyukov PTAL and let's then discuss.

I've [migrated](https://github.com/SpineEventEngine/mc-java/pull/173) `(compare_by)` comparator builder to the API from this branch. It has led to the following changes:

1. Similarly to variables, I've added Java elements to declare fields.
2. `JavaTypeName` hierarchy has been re-considered to cover parameterized types such as `Comparator<String>`. Now it looks like the [following](https://miro.com/app/board/uXjVLFfLAzU=/).

The reflective Java `Type` and Guava's `TypeToken` (and Kotlin's `typeOf()`) were insufficient for our use case. The main reasons are the following:

1. They require presence of the inspected types on the classpath, as they all rely on reflection and on Java `Type`. The generated messages usually are not present on the classpath of codegen. 
2. The returned string representation is not always printable directly to Java code. No support for non-resolved generics (like `String<T>`, which is valid field or variable type).
3. When used from Kotlin, it picks up **Java packages**, but **Kotlin generics**. So, when `TypeToken` is called for `List<String>` from Java, you get one result, from Kotlin - another. 

```kotlin
From Java: java.util.List<java.lang.String>
From Kotlin: java.util.List<? extends java.lang.String>
```

yevhenii-nadtochii:

java/src/main/kotlin/io/spine/protodata/java/FieldAccess.kt
```diff
@@ -0,0 +1,109 @@
...
+    kind: Cardinality,
+) : FieldConventions(name, kind) {
+
+    /**
+     * Constructs field access for the given [message] and [name].
+     */
+    internal constructor(
+        message: Expression<Message>,
+        name: String,
+        kind: Cardinality = Cardinality.CARDINALITY_SINGLE
```
alexander-yevsyukov, line 52: Please import enum items. They all have the prefix, so we'd know what they are about.

java/src/main/kotlin/io/spine/protodata/java/MethodCall.kt
```diff
@@ -0,0 +1,129 @@
...
+    public fun chainAddAll(field: String, value: Expression<*>): MethodCall<Message.Builder> =
+        fieldAccess(field).addAll(value)
+
+    private fun fieldAccess(fieldName: String) = FieldAccess(Expression(toCode()), fieldName)
+}
+
+/**
+ * The assumed reference to `this` when invoking a [MethodCall] within an instance.
+ */
+public class InstanceScope<T> : Expression<T>("")
```
alexander-yevsyukov, line 117: This would not work with the current implementation of `MethodCall` because it still has the dot notation in the string template (per my previous comments). 

I'm not suggesting to change the implementation of `MethodCall` now. Let's remove `InstanceScope` to avoid the misuse. We may return to revisiting `MethodCall` implementation later and then reintroduce the `InstanceScope` class.

java/src/main/kotlin/io/spine/protodata/java/MethodCalls.kt
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * The resulting expression always yields an instance of Guava `ImmutableMap`.
+ *
+ * @param expressions The expressions representing the entries.
+ * @param keyType The type of the keys in the map;
+ *   must be non-`null` if the map is not empty, may be `null` otherwise.
+ * @param valueType The type of the values in the map;
+ *   must be non-`null` if the map is not empty, may be `null` otherwise.
+ */
+public fun mapExpression(
```
alexander-yevsyukov, line 104: Can we have two overloads for this function? One would be without parameters representing construction of an empty map. The other one would be with non-empty parameter for `expressions` and non-`null` types for keys and values. 

This way we would not need to have an awkward check for non-`null` in nullable parameters.

alexander-yevsyukov (changes requested):
Nearly there. Please see my comments.

alexander-yevsyukov (approved):
LGTM

armiol (approved):

PR #292, opened 2024-11-21
Make message expressions covariant

yevhenii-nadtochii:
This PR introduces several small changes required to apply the updated Expressions API in Validation.

The changes are the following:

1. Make message-related expression extensions covariant, so that they could be used upon any children of `Message`.
2. Add `TypeName.javaClassName()` overload, which accept the type system.
3. Add `ClassName.call()` overload, which accepts a single argument only. There is a similar constructor of `MessageCall`.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine.protodata:protodata-api:0.68.0`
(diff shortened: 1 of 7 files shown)


yevhenii-nadtochii:

armiol (approved):

PR #294, opened 2024-11-27
Allow fetching a `ClassName` for enum messages

yevhenii-nadtochii:
This PR makes `TypeName.javaClassName()` extension work with enums as well.


alexander-yevsyukov (approved):
LGTM

PR #298, opened 2024-11-30
Errors and warnings

alexander-yevsyukov:
This PR provides initial API for dealing with compilation errors and warnings.

## Line number information
The PR introduces the `Span` type which provides coordinates for Protobuf declarations. Corresponding AST types were extended with the `span` fields.

Also, the helper class called `Coordinates` was introduced. It is similar to the `Documentation` class which helps with creation of `Doc` instances. 

Now both `Documentation` and `Coordinates` classes have the common `Locations` base. 
Instances of `Documentation` and `Coordinates` are created after `FileDescriptorProto` and are cached using the new `Cache` utility class.

The extension `GenericDescriptor.withSourceLines()` extension which tries to load a version of `GenericDescriptor` from a descriptor set file stored in resources. `Documentation` and `Coordinates` use this extension.

## Other notable changes
 * The caching of `SourceFile` instances was updated to use the new `Cache` class.
 * Added `Descriptor.nestedTypes()` extension function which filters the output of `getNestedTypes()` removing synthetic descriptors created for entry types of map fields.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/Compilation.kt
+++ b/api/src/main/kotlin/io/spine/protodata/Compilation.kt
@@ -59,7 +59,7 @@ public object Compilation {
      * The method returns the string printed to the console so that it could be also
      * put into logging output.
      *
-     * @param file The file in which causes the warning.
+     * @param file The file which causes the warning.
      * @param line The one-based number of the line with the questionable code.
      * @param column The one-based number of the column with the questionable code.
      * @param message The description of the warning.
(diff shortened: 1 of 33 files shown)


alexander-yevsyukov:
@armiol, PTAL.

armiol (approved):

api/src/main/kotlin/io/spine/protodata/Compilation.kt
```diff
@@ -0,0 +1,73 @@
...
+        exitProcess(ERROR_EXIT_CODE)
+    }
+
+    /**
+     * Prints the warning diagnostics to [System.out].
+     *
+     * The method returns the string printed to the console so that it could be also
+     * put into logging output.
+     *
+     * @param file The file in which causes the warning.
```
yevhenii-nadtochii, line 62:
```suggestion
     * @param file The file which causes the warning.
```

yevhenii-nadtochii (approved):

PR #299, opened 2024-12-04
`doc` and `span` properties for options

alexander-yevsyukov:
This PR adds `doc` and `span` properties for the `Option` type, and extends parsing of Protobuf descriptors for populating these properties.

## Changes in details
 * Extension functions called `options()` was added to the descriptor types. The functions return the options declared in the corresponding scope. 
 * The `ExtendableMessage.toList()` extension was hidden because the `options()` extensions now do the job.
 * `config` was also updated.


armiol (approved):
LGTM.

PR #305, opened 2024-12-13
Make `Compilation.Error` descend from `java.lang.Error`

alexander-yevsyukov:
This PR changes the parent type of `Compilation.Error` from `Exception` to `Error` so that plugin parts can terminate the compilation using the [recent update from CoreJava](https://github.com/SpineEventEngine/core-java/pull/1560).


armiol (approved):
LGTM.

PR #308, opened 2024-12-16
Rename displayed name for `bool`

yevhenii-nadtochii:
This PR makes `public val FieldType.name: String` extension return `bool` for boolean fields because Protobuf syntax says exactly `bool`.


yevhenii-nadtochii:

armiol (approved):

PR #309, opened 2024-12-19
Introduce `PipelineParameters` message type for creating pipelines

alexander-yevsyukov:
This PR introduces `PipelineParameters` message types and adopts it for passing the parameters to a `Pipeline`.

From time to time we needed one more parameter passed to a `Pipeline`. The latest one was the list of proto source files compiled by `protoc`. Adding a parameter involved a significant number of changes across both in production and test code. Wrapping all the `Pipeline` parameters into a parameter object addresses this issue. 

In addition to that, having a parameters type on a file system would help with debugging. Also, that allows to reduce the length of the command line for starting the ProtoData console app.

Details of this change and other related improvements are described in the sections below.

## The `cli-api` module was renamed to `params`
The module now contains the API for handling parameters and settings stored under `WorkingDiretory`.

## Command line parameters of the ProtoData console app
The CLI parameter `--params` was introduced replacing most of the existing parameters. Other remaining parameters are `--info` and `--debug` that set corresponding logging level when running the console app.

## Production and test modes of using the `PipelineParameters` type
Even though the `PipelineParameters` type declares validation constraints, it can be passed to a `Pipeline` as `@NonValidated`. The `Pipeline` class would do its best to keep going with partial parameters. Such an arrangement is needed for running plugin tests without specifying all the parameters.

Extension functions for building partial `PipelineParameters` instances were introduced in the `testlib` module:
 * `pipelineParams` — the block function which calls `buildPartial()` in the builder.
 * `withRequestFile`, `withSettingsDir`, `withRoots` — add corresponding parameters to the builder.

## `plugins` parameter of the `Pipeline` constructor
In the production mode the names of the plugin classes are passed via the `PipelineParameters.plugin_class_name` property. In tests, it is more convenient sometimes to pass just an instance of a plugin. This need is treated by the `plugins` parameter passed to the constructor of `Pipeline`.

## Removal of `CreateSettingsDirectory` task
The task was removed in favour of the `WorkingDirectory` helper object which deals with the files and directories needed for passing parameters and settings to ProtoData.

## `ProtoFileList` class
The purpose of this helper class is to locate absolute file paths via relative paths when reporting compilation errors or warnings. This PR adds it as the `compiledProtoFiles` property of a `TypeSystem`. More work around this class will be done in one of the following PRs.

##  `CodeGeneratorRequestExts.kt`
Sometimes test need a manually created instance of `CodeGeneratorRequest` message which is normally created by `protoc`. Creating a request requires many descriptors of dependencies of the compiled proto files. The `testlib` module introduces functions that list descriptors of types from the Protobuf library, as well as descriptors of the custom options provided by the Spine SDK.


api/src/main/kotlin/io/spine/protodata/protobuf/ProtoFileList.kt
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
```
armiol, line 2: We'll need to update the year to 2025 in `config` some time soon :)

api/src/main/kotlin/io/spine/protodata/protobuf/ProtoFileList.kt
```diff
@@ -0,0 +1,42 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.protodata.protobuf
+
+import java.io.File
+
+/**
+ * The list of absolute names of Protobuf files compiled by `protoc`.
```
armiol, line 32: Are these "files denoted by their absolute paths"?

api/src/main/kotlin/io/spine/protodata/settings/WithSettings.kt
```diff
@@ -27,31 +27,31 @@
 package io.spine.protodata.settings
 
 /**
- * A ProtoData component which obtains settings provided by the user.
+ * A ProtoData plugin component which can [load][loadSettings] its settings.
+ *
+ * It is the API user's responsibility to know the format of the settings and provide
+ * an appropriate class given s the parameter to the [loadSettings] method.
```
armiol, line 33: There's a typo in " given s the".

cli/src/test/kotlin/io/spine/protodata/cli/app/MainSpec.kt
```diff
@@ -103,12 +126,17 @@ class MainSpec {
...
-            ))
+            protoFile.addAll(
+                listOf(
+                    project.toProto(),
+                    testProto.toProto(),
+                    TestOptionsProto.getDescriptor().toProto(),
+                    AstProto.getDescriptor().toProto(),
+                    FileProto.getDescriptor().toProto(),
+                ) +
+                spineOptionProtos() +
```
armiol, line 137: This fragment is formatted weirdly. The [same piece in `LoggingLevelSpec`](https://github.com/SpineEventEngine/ProtoData/pull/309/files#diff-f448e9b8864c0dcd30ed0e3ec592ded83a7435e33aec10982366d9eb209d0bebR109) looks better to me.

params/src/main/proto/spine/protodata/params/pipeline_parameters.proto
```diff
@@ -0,0 +1,107 @@
...
+
+    // The path to the binary file containing a serialized instance of
+    // `com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest` received by
+    // ProtoData `protoc` plugin.
+    //
+    File request = 4 [(required) = true];
+
+    // Paths to directories with the source files to be processed.
+    //
+    // Skip this parameter if there is no initial source to modify.
```
armiol, line 74: It's `[(required) = true]`, it cannot be skipped.

params/src/main/proto/spine/protodata/params/pipeline_parameters.proto
```diff
@@ -0,0 +1,107 @@
...
+    // Please see the documentation of the `source_root` parameter for the rules of matching
+    // source and target root directories.
+    //
+    repeated Directory target_root = 6 [(required) = true];
+
+    // The user classpath which contains all plugin classes specified in the `user_class_name`
+    // property, as well as their dependencies, which are not included as a part of
+    // the ProtoData library.
+    //
+    // May be one path to a JAR, a ZIP, or a directory, or several entries of these kinds.
```
armiol, line 102: "...a single path to a JAR, ..."

armiol (approved):
@alexander-yevsyukov LGTM except for a few documentation-related comments. Please see them in-line.

PR #311, opened 2024-12-24
Extend `MethodCall` with `put` methods

yevhenii-nadtochii:
This changeset does the following:

1. Extends `MethodCall` with `chainPut()` and `chainPutAll()` methods.
2. Changes the parameters order of `mapExpression()` method. Looks like declaration of the map key and value types should go before the actual entry list. We are used to reading declaration statements in this order.
3. Introduces `LongLiteral` that appends "L" ending to produced Java code.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine.protodata:protodata-api:0.80.4`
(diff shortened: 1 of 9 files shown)


yevhenii-nadtochii:

armiol (approved):

PR #316, opened 2025-01-16
Take the latest `tool-base` and bring a handful of improvements

yevhenii-nadtochii:
This PR does the following:

1. Takes the latest `tool-base` to pick up [tool-base #111](https://github.com/SpineEventEngine/tool-base/pull/111).
2. Makes the whole `Expression` hierarchy covariant. Previously, only some child members declared it.
3. `FieldDeclaration` is actually `Statement`.

### Remove nullability of `Member` properties

I'm not 100% sure about this change, but personally don't foresee any problems.

I suggest moving null safety checks from the usage site to the declaration site of `Member.context` and `typeSystem` properties. Making them non-nullable is desired for the following reasons:

1. I have already met the third `JavaRenderer` that has such a line `public fun typeSystem(): TypeSystem = typeSystem!!`.
2. `!!` leads to a naked NPE while ISE from the declaration site would bring an error message.
3. Anyway, each line of code within renderers (and, probably, members) is meant to be executed during the codegen. 
4. We have `select()` method in `Member`, and it just throws when used without the injected codegen context, but could have returned `null` as properties do.

### DIFF
--- a/cli/src/main/kotlin/io/spine/protodata/cli/app/Main.kt
+++ b/cli/src/main/kotlin/io/spine/protodata/cli/app/Main.kt
@@ -53,7 +53,7 @@ import kotlin.system.exitProcess
  * Launches the CLI application.
  *
  * When the application is done exists the process with the code `0`.
- * If an unhandled error occurs, exits the process with the code `1`.
+ * If an unhandled error occurs, exits the process with the code `-1`.
  */
 @Suppress(
     "TooGenericExceptionCaught", // We do want the most generic type thrown.
@@ -72,7 +72,7 @@ public fun main(args: Array<String>) {
             e.printStackTrace(this)
             println("```")
         }
-        exitProcess(1)
+        exitProcess(-1)
     }
 }
 
--- a/cli/src/main/kotlin/io/spine/protodata/cli/app/Main.kt
+++ b/cli/src/main/kotlin/io/spine/protodata/cli/app/Main.kt
@@ -53,7 +53,7 @@ import kotlin.system.exitProcess
  * Launches the CLI application.
  *
  * When the application is done exists the process with the code `0`.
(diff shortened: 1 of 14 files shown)


yevhenii-nadtochii:

armiol:
@yevhenii-nadtochii The reason for returning 255 and not 1 is exactly the commonality of the latter. We want to distinguish the case in which we made a deliberate choice to stop the machine and the cases in which some third-party code (or JVM itself) has exited due to their particular reasons.

java/src/main/kotlin/io/spine/protodata/java/ProtobufExpressions.kt
```diff
@@ -58,11 +58,11 @@ public fun Expression<*>.packToAny(): Expression<ProtoAny> {
 /**
  * Obtains a [FieldAccess] to the [field] of this message.
  */
-public fun Expression<out Message>.field(field: Field): FieldAccess =
+public fun Expression<Message>.field(field: Field): FieldAccess =
```
armiol, line 61: As long as it allows you to use this code in Validation, looks logical.

cli/src/main/kotlin/io/spine/protodata/cli/app/Main.kt
```diff
@@ -72,7 +72,7 @@ public fun main(args: Array<String>) {
             e.printStackTrace(this)
             println("```")
         }
-        exitProcess(-1)
```
armiol, line 75: See my comment on the PR level.

armiol (changes requested):
@yevhenii-nadtochii LGTM except for the exit code change.

yevhenii-nadtochii:
@armiol PTAL

armiol (approved):

alexander-yevsyukov (approved):
LGTM

PR #318, opened 2025-02-04
Make `Field` always contain the declared options

yevhenii-nadtochii:
This PR makes `field.optionList` always contain the declared options of the field. Previously, `field` lacked them when it arrived as a part of `FieldOptionDiscovered` event.

Also, this PR brings a bunch of changes to Expression API:

1. Adds `CodeBlock` Java element to represent arbitrary blocks of code.
2. Adds `MemberDeclaration` to represent class-level declarations. `FieldDeclaration` and `MethodDeclaration` are inheritors. As for now, `MethodDeclaration` accepts just `code` to pass the whole method as text.
3. `This()` expression now accepts `explicit` flag. `MethodCall` now omits `$scope.` when `scope.isBlank()`, making possible implicit method invocations.
4. Adds a convenient shortcut `Expression<Message>.toBuilder(): MethodCall`.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
+++ b/api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
@@ -84,7 +84,7 @@ import kotlin.reflect.KClass
 public fun FieldDescriptor.name(): FieldName = fieldName { value = name }
 
 /**
- * Converts this field descriptor into a [Field] with options.
+ * Converts this field descriptor into a [Field].
  */
 public fun FieldDescriptor.toField(): Field =
     field {
--- a/api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
+++ b/api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
@@ -94,7 +94,7 @@ public fun FieldDescriptor.toField(): Field =
         // New `FieldType` and `group_name` API.
         type = toFieldType()
         declaringType = declaredIn
-        number = number
+        number = this@toField.number
         orderOfDeclaration = index
         realContainingOneof?.let {
             enclosingOneof = it.name()
--- a/api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
+++ b/api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
@@ -68,7 +68,6 @@ import io.spine.protodata.ast.PrimitiveType.TYPE_UINT64
 import io.spine.protodata.ast.Type
 import io.spine.protodata.ast.TypeName
 import io.spine.protodata.ast.coordinates
-import io.spine.protodata.ast.copy
 import io.spine.protodata.ast.documentation
 import io.spine.protodata.ast.field
 import io.spine.protodata.ast.fieldName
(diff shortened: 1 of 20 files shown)


yevhenii-nadtochii:

java/src/main/kotlin/io/spine/protodata/java/MethodDeclaration.kt
```diff
@@ -0,0 +1,45 @@
...
+ * An example usage:
+ *
+ * ```
+ * val validate = MethodDeclaration(
+ *     """
+ *     public java.util.Optional<io.spine.validate.ValidationError> validate() {
+ *         var noParent = io.spine.base.FieldPath.getDefaultInstance();
+ *         return validate(noParent);
+ *     }
+ *     """.trimIndent()
```
armiol, line 41: What is the outcome of your investigation regarding `trimIndent()` and `trimMargin()`?

armiol (commented):
@yevhenii-nadtochii please see my question.

api/src/main/kotlin/io/spine/protodata/protobuf/FieldDescriptorExts.kt
```diff
@@ -84,45 +83,25 @@ import kotlin.reflect.KClass
...
-        name = desc.name()
+        name = name()
         // New `FieldType` and `group_name` API.
-        type = desc.toFieldType()
+        type = toFieldType()
         declaringType = declaredIn
-        number = desc.number
-        orderOfDeclaration = desc.index
-        desc.realContainingOneof?.let {
+        number = number
```
alexander-yevsyukov, line 97: This `number = number` looks suspicious. Perhaps a reference to the context would make it clearer.

java/src/main/kotlin/io/spine/protodata/java/MethodDeclaration.kt
yevhenii-nadtochii: PSI throws when the passed method declaration has leading whitespaces. I've left a [note](https://github.com/SpineEventEngine/validation/pull/188/files#diff-f2902148e4d7730176b75979fb22fb1e06e360d096b01eb46dd6af020e7dbac8R190), and have used the simplest `trim()`.

alexander-yevsyukov (commented):
Please see my comment. Otherwise, LGTM.

alexander-yevsyukov (approved):
LGTM, but please introduce the var.

java/src/main/kotlin/io/spine/protodata/java/MethodDeclaration.kt
armiol: @yevhenii-nadtochii The question is boiled down to this: will this "example usage" work?

yevhenii-nadtochii:
@armiol PTAL

java/src/main/kotlin/io/spine/protodata/java/JavaElement.kt
```diff
@@ -39,7 +39,28 @@ public interface JavaElement : CodeElement<Java>
...
+ * common indent. This effectively causes an invocation of [trimIndent] to behave
+ * as a no-op.
+ *
+ * Concrete implementations of [AnElement] may not accept [code] on constructor,
+ * requiring the user to pass more specific parts of the code. In such cases,
+ * it is a responsibility of the class to assemble the correct final [code].
+ * Specific subclasses may also accept [code] that allows leading whitespaces;
+ * if so, they should document this behavior explicitly.
+ *
+ * @param code Arbitrary Java code without leading whitespaces.
```
armiol, line 63: As discussed, let's add some code examples.

And the second point discussed: please have this type referenced whenever we have a code sample with `trimIndent()`, so that the users of this API could know it's not _that_ simple in all the cases.

armiol (changes requested):

yevhenii-nadtochii:
@armiol PTAL

I've removed trimming from Expression API.

armiol (commented):
@yevhenii-nadtochii please see my comment.

java/src/main/kotlin/io/spine/protodata/java/CodeBlock.kt
```diff
@@ -0,0 +1,58 @@
...
+ * CodeBlock(
+ *     """
+ *     // Create a violation if the field value equals to "hello".
+ *     if (fieldValue.equals("hello")) {
+ *         var fieldPath = getFieldPath();
+ *         var violation = createViolation(fieldPath);
+ *         violations.add(violation);
+ *     }
+ *     return violations;
+ *     """.trimIndent()
```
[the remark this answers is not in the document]
yevhenii-nadtochii: No. Though, I never remove automatically appended `trimIndent()` until I really need to.

Another point is that the following:

```
// Create a violation if the field value equals to "hello".
if (fieldValue.equals("hello")) {
    var fieldPath = getFieldPath();
    var violation = createViolation(fieldPath);
     violations.add(violation);
}
return violations;
```

is more expected compared to this:

```
    // Create a violation if the field value equals to "hello".
    if (fieldValue.equals("hello")) {
        var fieldPath = getFieldPath();
        var violation = createViolation(fieldPath);
         violations.add(violation);
    }
    return violations;
```

Both options may be valid for the user. Maybe they want to insert it into the block with specific nesting. Or maybe they don't know the future nesting, and want just an arbitrary set of statements. In the latter way, indent is not needed because it is unknown in advance (4, 8, 12 spaces?).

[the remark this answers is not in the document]
armiol: @yevhenii-nadtochii the whole point of this lengthy discussion is to allow end-users of this API to not have to think about these details. As much as possible. Now, it turns out we need to keep the cryptic things like `trimIndent()`.

Let's discuss this whole thing one more time on Monday.

armiol (approved):

PR #321, opened 2025-02-14
Allow instantiation of `FieldDeclaration`

yevhenii-nadtochii:
This PR merges inheritors of `FieldDeclaration` into a single class, allowing instantiation of the mentioned class from arbitrary code. The merged classes are now represented by secondary constructors.

We need this change to perform conversion of JavaPoet's `FieldSpec` without digging into details about what kind of field it bears.


yevhenii-nadtochii:

api/src/main/kotlin/io/spine/protodata/CompilationErrors.kt
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * If the code is run [under tests][Tests] the method throws [Compilation.Error].
+ *
+ * @param file The file in which the error occurred.
+ * @param span The location span of the erroneous Protobuf declaration.
+ * @param message The description of what went wrong.
+ *
+ * @throws Compilation.Error exception when called under tests.
+ */
+public fun compilationError(file: File, span: Span, message: () -> String): Nothing =
```
alexander-yevsyukov, line 49: Can we have it as another function under the `Compilation` object, please?

alexander-yevsyukov: Or, if we insist on having such things as top level functions, rather than mathods of the `Compilation` object, this should be discussed first.

alexander-yevsyukov: The reason we have the `Compilation` object is potential need to cache something internal things. Or, do not terminate immediately on the first error and count them until terminating the compilation.

Such things are more natural to do under a class or an object.

alexander-yevsyukov (changes requested):
Please see my comment.

java/src/main/kotlin/io/spine/protodata/java/Fields.kt
```diff
@@ -35,8 +35,15 @@ import io.spine.tools.psi.java.Environment.elementFactory
...
+ *
+ * Note: the class is intentionally not `sealed`, allowing instantiation
+ * with arbitrary [code]. While not recommended for general use, this can
+ * be helpful, e.g., when converting JavaPoet or PSI fields to [FieldDeclaration].
+ *
+ * @see InitField
+ * @see DeclField
  */
-public sealed class FieldDeclaration<T>(
+public open class FieldDeclaration<T>(
```
armiol, line 46: How does this `sealed`->`open` change correlate with the portion of the PR description? Namely:
> We can use this class to directly pass code when converting JavaPoet or PSI types to ProtoData counterparts, significantly simplifying the conversion

What I read here is as follows: "We need to instantiate the objects of this type by passing the code obtained via JavaPoet or PSI, directly." But in the scope of other modules (such as Validation), even a `sealed` version of this class could be used to create instances. Be it with the code obtained from JavaPoet, or from anywhere else.

Therefore, I say something is missing in the PR description. I suspect what you wanted was to create descendant types for `FieldDeclaration`. And if it is so, you should a) write about it explicitly, and b) consider having some tests in this project, so that we are sure this type remains `open`.

armiol (changes requested):
@yevhenii-nadtochii please see my comment.

yevhenii-nadtochii:

armiol (approved):

alexander-yevsyukov (approved):
LGTM

PR #325, opened 2025-02-17
Finding via absolute and relative paths

alexander-yevsyukov:
This PR adds the `Member.findSource()` extension which allows to find an instance of `ProtobufSourceFile` via relative or full file path. This, in turn, fixes the issue `Member.findHeader()` extension function which used to find the header only via a relative path, which used to be an ID of corresponding entity states. Now both absolute and relative paths work.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/context/Member.kt
+++ b/api/src/main/kotlin/io/spine/protodata/context/Member.kt
@@ -160,10 +160,23 @@ protected constructor(
 }
 
 /**
- * Obtains the header of the proto file with the given [path].
+ * Obtains the proto source file with the given [path].
+ *
+ * @param path The path to the source file which could be relative or absolute.
  */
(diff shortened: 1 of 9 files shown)


api/src/main/kotlin/io/spine/protodata/context/Member.kt
```diff
@@ -159,11 +159,24 @@ protected constructor(
     }
 }
 
+/**
+ * Obtains the proto source file with the given [path].
+ *
+ * @param path The path to the source file which could be relative or absolute.
+ */
+public fun Member<*>.findSource(path: File): ProtobufSourceFile? {
+    return (this as Querying).select<ProtobufSourceFile>().findById(path)
```
yevhenii-nadtochii, line 168: It is a single line expression, you can save on braces.

And please, mention when it is OK for this function to return `null`. The same suggestion is to the function below.

yevhenii-nadtochii (approved):
LGTM with a comment.

PR #329, opened 2025-02-20
Fix full paths in `ProtoDeclaration` implementions

alexander-yevsyukov:
This PR fixes setting absolute file paths in `MessageType`, `EnumType`, and `Service`. Previously, only events related to the declarations and file headers had full paths.

### DIFF
--- a/api/src/test/proto/protodata/given/stub_declaration.proto
+++ b/api/src/test/proto/protodata/given/stub_declaration.proto
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024, TeamDev. All rights reserved.
+ * Copyright 2025, TeamDev. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


api/src/test/proto/protodata/given/stub_declaration.proto
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
```
armiol, line 2: It's 2025.

cli/src/test/proto/spine/protodata/cli/test/test.proto
```diff
@@ -14,6 +14,10 @@ option java_multiple_files = true;
 import "google/protobuf/timestamp.proto";
 import "spine/protodata/ast.proto";
 
+message Project {
```
armiol, line 17: Couldn't quickly find if it is used. Just saying. Looks a bit odd.

armiol (approved):
LGTM with some remarks.

PR #331, opened 2025-02-25
Bunch of enhancements from `validation`

yevhenii-nadtochii:
This PR bring a handful of enhancements suggested in `validation`.

The changes are the following:

1. `FieldRef` proto type to uniquely identify a field within a compilation process.
2. `Policy.typeSystem` is non-nullable now.
3. Added throwing method counterparts for `TypeSystem`. They throw instead of returning `null` when the requested declaration is not found. Also, they don't return a file. We shouldn't need a file if we already know that the declaration is for sure present.
4. Added `clean` task to `RunBuild` execution, which is used for integration tests. We may run into non-obvious problems when `build` directories under `tests` are stale.
5. Added `MessageType.findField()` extension.
6. Added shortcuts to create `MethodCall` upon `Expression`.

### DIFF
--- a/java/src/main/kotlin/io/spine/protodata/java/MessageTypeExts.kt
+++ b/java/src/main/kotlin/io/spine/protodata/java/MessageTypeExts.kt
@@ -95,7 +95,7 @@ public fun MessageType.javaClass(typeSystem: TypeSystem): Class<*>? {
 }
 
 /**
- * Looks up for a field with the given [name] in this [MessageType].
+ * Looks for the field with the given [name] in this [MessageType].
  *
  * @return the found [Field], or `null` if this [MessageType] does not have such a field.
  */
--- a/api/src/main/proto/spine/protodata/ast.proto
+++ b/api/src/main/proto/spine/protodata/ast.proto
@@ -353,7 +353,7 @@ message Field {
 // Uniquely identifies a field declaration within a compilation process.
 message FieldRef {
 
-    // The name of the containing type.
+    // The name of the declaring type.
     TypeName type = 1;
 
     // The name of the field.


api/src/main/kotlin/io/spine/protodata/type/TypeSystem.kt
```diff
@@ -185,3 +185,30 @@ public fun TypeSystem.resolve(fieldPath: FieldPath, message: MessageType): Field
     val nextMessage = nextMessageInfo.first
     return resolve(remainingPath, nextMessage)
 }
+
+/**
+ * Obtains a message type by its name.
+ *
+ * @throws IllegalArgumentException if the requested type is unknown for this [TypeSystem].
```
armiol, line 192: All extensions throw more or less the same exception, but for some reason it is only documented here. Please adjust other descriptions accordingly.

api/src/main/proto/spine/protodata/references.proto
```diff
@@ -0,0 +1,48 @@
...
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.protodata.ast";
+option java_outer_classname = "ReferencesProto";
+option java_multiple_files = true;
+
+import "spine/protodata/ast.proto";
+
+// Uniquely identifies a field declaration within a compilation process.
+message FieldRef {
```
armiol, line 41: To me, this is a part of `ast.proto`. You even have the same Java package for it.

java/src/main/kotlin/io/spine/protodata/java/MessageTypeExts.kt
```diff
@@ -92,3 +93,11 @@ public fun MessageType.javaClass(typeSystem: TypeSystem): Class<*>? {
     val name = javaClassName(typeSystem)
     return name.javaClass()
 }
+
+/**
+ * Looks up for a field with the given [name] in this [MessageType].
```
armiol, line 98: Either "looks up the field" or "looks for the field".

api/src/main/proto/spine/protodata/references.proto
```diff
@@ -0,0 +1,48 @@
...
+
+    // The name of the containing type.
```
alexander-yevsyukov, line 43: I'd suggest to use the word "declaring" because with "containing" it looks like it's the type of the field.

api/src/main/kotlin/io/spine/protodata/ast/References.kt
armiol, line 1: `FieldExts.kt` is already not too homogeneous. So I would put `Field.ref` there, too.

api/src/main/proto/spine/protodata/references.proto
```diff
@@ -0,0 +1,48 @@
...
+    protodata.TypeName type = 1;
+
+    // The name of the field.
+    protodata.FieldName name = 2;
```
alexander-yevsyukov, line 47: Why do we need the `protodata` prefix here and above? We're in this package already.

armiol (changes requested):
@yevhenii-nadtochii please see my comments.

alexander-yevsyukov (changes requested):
Please see my comments.

yevhenii-nadtochii:

armiol (approved):

alexander-yevsyukov (approved):
LGTM

alexander-yevsyukov (approved):
LGTM

PR #333, opened 2025-02-27
Add extensions to find a field option by its type

yevhenii-nadtochii:
This PR adds two extensions to find a field option by its type:

1. `Field.findOption()`, which returns a nullable result.
2. `Field.option()`, which throws ISE if the option not found.

Additionally, it adds extensions to get a message `Descriptor` from the message `Class`:

1. `Class<out Message>.descriptor: Descriptor`  – for Java classes.
2. `KClass<out Message>.descriptor: Descriptor` - for Kotlin classes.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
+++ b/api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
@@ -33,10 +33,11 @@ import kotlin.reflect.KClass
 /**
  * Returns a Protobuf descriptor for this message [Class].
  */
-public fun Class<out Message>.getDescriptor(): Descriptor =
-    getDeclaredMethod("getDescriptor").invoke(null) as Descriptor
+public val Class<out Message>.descriptor: Descriptor
+    get() = getDeclaredMethod("getDescriptor").invoke(null) as Descriptor
 
 /**
  * Returns a Protobuf descriptor for this message [KClass].
  */
-public fun KClass<out Message>.getDescriptor(): Descriptor = java.getDescriptor()
+public val KClass<out Message>.descriptor: Descriptor
+    get() = java.descriptor
--- a/api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
+++ b/api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025, TeamDev. All rights reserved.
+ *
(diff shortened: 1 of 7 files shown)


api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Returns a Protobuf descriptor for this message [Class].
+ */
+public fun Class<out Message>.getDescriptor(): Descriptor =
+    getDeclaredMethod("getDescriptor").invoke(null) as Descriptor
+
+/**
+ * Returns a Protobuf descriptor for this message [KClass].
+ */
+public fun KClass<out Message>.getDescriptor(): Descriptor = java.getDescriptor()
```
yevhenii-nadtochii, line 42: They also [come](https://github.com/SpineEventEngine/validation/pull/192/files#diff-ec2ae3a3f10ff63a7085eccf7a47cbda13c38a13da75b258647d3f37857c09caR51) from `validation`. But they look too generic for the `validation` library.

These are also [similar](https://github.com/SpineEventEngine/validation/pull/192/files#diff-49793842ebfeaf401f8cd6048d0d4a64b9edbaa9bfb40cb0f35920ceafe2ae04R35) extensions to return `Message.Builder`, but they are tests-only. As for now, I can't say how they can be used in `main` sources.

yevhenii-nadtochii:

armiol:
@yevhenii-nadtochii Is this still a draft?

yevhenii-nadtochii:
> @yevhenii-nadtochii Is this still a draft?

No, it is ready. Fixed the state.

api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
armiol, line 42: Maybe, `descriptor()` is enough? Why `get...()`?

armiol (approved):
@yevhenii-nadtochii LGTM with a suggestion, from my side.

api/src/main/kotlin/io/spine/protodata/protobuf/MessageClass.kt
yevhenii-nadtochii: I wanted it to look more like the original static method: `Message.getDescriptor()`, which we actually invoke through the reflection.

armiol: @yevhenii-nadtochii I don't think we are that interested in vintage.

PR #335, opened 2025-03-05
Bump Gradle, Kotlin, and other dependencies

alexander-yevsyukov:
This PR updates `config` brining new major version of Gradle and Kotlin. Also, Base, McJava, and Validation build with new versions of Gradle and Kotlin are also updated as dependencies.

The deprecations in Gradle API both in build scripts and ProtoData Gradle Plugin implementation were also addressed.

## Other notable changes
 * Changed escaping in `(pattern).regex` values to match latest updates in Validation.
 * `LaunchTask` locator was renamed to `ProtoDataTask` to better reflect the nature of the task to locate.


alexander-yevsyukov:
@armiol, в первом приближении билдится. Нужно ещё несколько deprecations подправить, но их немного.

buildSrc/src/main/kotlin/io/spine/dependency/local/ToolBase.kt
```diff
@@ -34,7 +34,7 @@ package io.spine.dependency.local
 @Suppress("ConstPropertyName", "unused")
 object ToolBase {
     const val group = Spine.toolsGroup
-    const val version = "2.0.0-SNAPSHOT.246"
+    const val version = "2.0.0-SNAPSHOT.244"
```
yevhenii-nadtochii, line 37:
```suggestion
    const val version = "2.0.0-SNAPSHOT.300"
```

yevhenii-nadtochii (approved):

api/src/main/proto/spine/protodata/ast.proto
```diff
@@ -74,7 +74,7 @@ message TypeName {
     //
     // Must not include a trailing slash ("/").
     //
-    string type_url_prefix = 4 [(pattern).regex = "^.*[^/]$"];
+    string type_url_prefix = 4 [(pattern).regex = "^.*[^\\/]$"];
```
armiol, line 77: I am still not sure whether `\\` is what we need. Does it work properly without the backslashes?

api/src/main/proto/spine/protodata/ast.proto
```diff
@@ -405,7 +405,7 @@ message ServiceName {
     //
     // Must not include a training slash ("/").
     //
-    string type_url_prefix = 3 [(pattern).regex = "^.*[^/]$"];
+    string type_url_prefix = 3 [(pattern).regex = "^.*[^\\/]$"];
```
armiol, line 408: Same as above.

buildSrc/src/main/kotlin/io/spine/dependency/test/KotlinCompileTesting.kt
```diff
@@ -0,0 +1,40 @@
...
+package io.spine.dependency.test
+
+/**
+ * A library for in-process compilation of Kotlin and Java code compilation.
+ *
+ * @see <a href="https://github.com/tschuchortdev/kotlin-compile-testing">GitHub repo</a>
+ */
+@Suppress("unused", "ConstPropertyName")
+object KotlinCompileTesting {
+    private const val version = "1.5.0" // Compatible with Kotlin Compiler 1.8.22.
```
armiol, line 36: I think it is safe to delete this dependency now, as we migrated to its "competitor".

buildSrc/src/main/kotlin/module.gradle.kts
```diff
@@ -24,6 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+import Module_gradle.Module
```
armiol, line 27: Please kill this one.

armiol (changes requested):
@alexander-yevsyukov LGTM in general. However, please see my comments.

api/src/main/proto/spine/protodata/ast.proto
yevhenii-nadtochii: >> Regarding (5): Java Regex treats \ as en escape for /, which indeed does not even have to be escaped. So, [^\\/] is the same as [^/].

https://github.com/SpineEventEngine/validation/pull/195#issuecomment-2700946186

yevhenii-nadtochii: They make no difference.

api/src/main/proto/spine/protodata/ast.proto
alexander-yevsyukov: It works without.

buildSrc/src/main/kotlin/io/spine/dependency/test/KotlinCompileTesting.kt
alexander-yevsyukov: I'll update the coordinates to those we used in McJava.

armiol (approved):

PR #346, opened 2025-04-14
Fill in the `span` property of `OneofGroup`

yevhenii-nadtochii:
This PR re-uses the `OneofDescriptor.toOneOfGroup()` method to create `OneofGroup` instance. Created this way, the `OneofGroup.span` field will be correctly populated. Thus, emitted compilation errors will point to the correct file and position.

Other changes are the following:

1. Added `OneofGroup.qualifiedName` extension, as we have for `Field`.
2. Added `TypeName.findJavaClassName()` counterpart that returns a nullable result instead of throwing.
3. Took the latest `core-java` to pick up the latest changes in `spine-time`.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/ast/Coordinates.kt
+++ b/api/src/main/kotlin/io/spine/protodata/ast/Coordinates.kt
@@ -161,7 +161,7 @@ private fun Location.toSpan(): Span {
 }
 
 /**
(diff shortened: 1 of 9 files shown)


yevhenii-nadtochii:

alexander-yevsyukov (approved):
LGTM

PR #353, opened 2025-05-16
Prepend indent for user-provided multi-line compilation messages

yevhenii-nadtochii:
This PR updates `Compilation` object, so that it prints multi-line error and warning messages, indenting each following line (after the first) with the length of the leading prefix.

An example of a real compilation error from `validation` using this version of ProtoData:

```text
> Task :java-tests:validating:launchTestFixturesProtoData FAILED
e: file:///Users/yevhenii/Projects/Spine/validation-master/java-tests/validating/src/testFixtures/proto/spine/test/tools/validate/numbers.proto:39:5: The `(range)` option could not parse the passed range value.
   Value: `[5.0 .. +1.0]`.
   Target field: `spine.test.tools.validate.Probability.value`.
   Reason: the lower bound `5.0` must be less than the upper bound `1.0`.
   Examples of the correct ranges: `(-5..5]`, `[0 .. 10)`.
```

## Concatenation for `String` expressions

Additionally, this PR moves operators for `String` expression concatenation introduced in `validation`.

### DIFF
--- a/api/src/main/kotlin/io/spine/protodata/Compilation.kt
+++ b/api/src/main/kotlin/io/spine/protodata/Compilation.kt
@@ -33,10 +33,12 @@ import io.spine.protodata.Compilation.ERROR_EXIT_CODE
 import io.spine.protodata.Compilation.error
 import io.spine.protodata.ast.Span
 import io.spine.protodata.ast.toJava
+import io.spine.string.Indent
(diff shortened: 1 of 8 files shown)


yevhenii-nadtochii:

alexander-yevsyukov (approved):
LGTM

